PR #13, opened 2024-06-27
Refactor: refactor + tests for opcode enum

nerodesu017:
### Pull Request Overview

This pull request refactors so that ~an enum is used for opcodes instead of their raw byte representation~ we use constant values in order to improve the readability of the code. This pull request also adds test cases for the above implementation.

~Also, the usage of `panic` is correct, as we would exit out at the validation point when an instruction we haven't implemented yet is encountered. It is also used for correctly assessing the outputs of the tests.~

### Testing Strategy

This pull request was tested locally using the test cases provided.

### TODO or Help Wanted

This pull request still needs an opinion if it is the right approach.

### Formatting

- [x] Ran `cargo fmt`
- [x] Ran `cargo check`
- [x] Ran `cargo build`

### Github Issue

N/A

### Author

Signed-off-by: Popescu Adrian <adrian.popescu@oxidos.io>


src/execution/mod.rs
```diff
@@ -208,17 +211,33 @@ impl<'b> RuntimeInstance<'b> {
...
+                    trace!("Instruction: i32.mul [{v1} {v2}] -> [{res}]");
+                    stack.push_value(res.into());
                 }
-                other => {
-                    trace!("Unknown instruction {other:#x}, skipping..");
+                Op1Byte::FCInstructions => {
+                    let second_byte = wasm.read_u8().unwrap_validated();
+                    let parsed_2byte_instr =
+                        Op2Byte::new(u16::from_be_bytes([first_byte, second_byte]));
+                    match parsed_2byte_instr {
```
wucke13, line 230: I'm not a fan of the nested match statements

wucke13:
Here is another Idea I quickly stitched together; we should also get @florianhartung feedback on that as he dismissed this idea earlier, IIRC.

```rust
mod opcodes {
    const NOP: u8 = 0x01;
    const End: u8 = 0x0B;
    const LocalGet: u8 = 0x20;
    const LocalSet: u8 = 0x21;
    const GlobalGet: u8 = 0x23;
    const GlobalSet: u8 = 0x24;
    const I32Load: u8 = 0x28;
    const I32Store: u8 = 0x36;
    const I32Const: u8 = 0x41;
    const I32Add: u8 = 0x6A;
    const I32Mul: u8 = 0x6C;
    const FCInstructions: u8 = 0xF;
}

mod fc_opcodes {
    const I32TruncSatF32s: u8 = 0x00;
}

struct Interpreter<'a> {
    bytecode: &'a [u8],
    pc: usize,
    stack: Vec<u8>,
}

impl<'a> Interpreter<'a> {
    fn run(&mut self) -> Result<(), Box<dyn std::error::Error>> {
        use fc_opcodes::*;
        use opcodes::*;

        let Self {
            pc,
            bytecode,
            stack,
        } = self;

        match &bytecode[*pc..] {
            [NOP, ..] => {
                *pc += 1;
            }
            [LocalSet, x, ..] => {
                stack.push(*x);
                *pc += 2;
            }
            [FCInstructions, I32TruncSatF32s, ..] => {
                // do some work
                *pc += 2;
            }
            _ => {
                todo!("something left to be implemented");
            }
        }

        Ok(())
    }
}
```

src/execution/mod.rs
nerodesu017: Okay, then how do you say we handle the fact that some instructions are 2 bytes? In that case we have to always read a u16. If there's another better way of doing this, I'm all ears.

nerodesu017:
Interesting way of doing this, I'd refactor it this way if Florian and Alex agree.

valexandru:
Looks clean to me, the only downside that I can find would be the refactoring of the code for this.

wucke13:
Thank you for the comments! There is two downsides to my way, it contains a variable amount of decisions (possibly hindering optimization by the Compiler), and the const stuff is rather elaborate. Let's wait for @florianhartung feedback before we continue this approach.

src/execution/mod.rs
```diff
@@ -219,56 +215,53 @@ where
...
                     stack.push_value(res.into());
                 }
-                // i32.mul: [i32 i32] -> [i32]
-                0x6C => {
+                I32_MUL => {
                     let v1: i32 = stack.pop_value(ValType::NumType(NumType::I32)).into();
                     let v2: i32 = stack.pop_value(ValType::NumType(NumType::I32)).into();
                     let res = v1.wrapping_mul(v2);
 
+                    trace!("Instruction: i32.mul [{v1} {v2}] -> [{res}]");
```
valexandru, line 257: Is this duplicated on purpose?

src/execution/mod.rs
```diff
@@ -129,47 +129,44 @@ where
...
+        use crate::core::reader::types::opcode::*;
         loop {
             // call the instruction hook
             #[cfg(feature = "hooks")]
             H::instruction_hook(self);
 
-            match wasm.read_u8().unwrap_validated() {
-                // end
-                0x0B => {
+            let first_byte = wasm.read_u8().unwrap_validated();
```
valexandru, line 138: I see the value in making a variable with the first byte, but maybe also rename it to something more suggestive: first_instr_byte

tests/instructions/multibyte/fc_extensions.rs
```diff
@@ -0,0 +1,41 @@
+/// A function that tests the FC multibyte instructions
+/// https://pengowray.github.io/wasm-ops/#fc
+///
+/// At the moment, it dies on validation, as the f32 data type is not implemented
+#[ignore] // we don't support multibyte instructions yet
+#[test_log::test]
+#[should_panic(expected = "i32.trunc_sat_f32_s NOT implemented")]
```
valexandru, line 7: Wouldn't this crash since there is no FC operation implemented?

nerodesu017: Well, that's why I added the `#[ignore]` attribute

valexandru (approved):

florianhartung (approved):
LGTM

PR #23, opened 2024-07-04
refactor: magic bytes to constant op codes

nerodesu017:
### Pull Request Overview

This pull request changes the way we handle opcodes: from using magic bytes to using constants.

### Testing Strategy

This pull request was tested locally.

### TODO or Help Wanted

N/A

### Formatting

- [x] Ran `cargo fmt`
- [x] Ran `cargo check`
- [x] Ran `cargo build`
- [ ] Ran `nix fmt`
- [ ] Ran `treefmt`

### Github Issue

N/A

### Author

Signed-off-by: Popescu Adrian <adrian.popescu@oxidos.io>

### DIFF
--- a/src/core/reader/types/values.rs
+++ b/src/core/reader/types/values.rs
@@ -14,14 +14,19 @@ use crate::{Error, Result};
 impl WasmReader<'_> {
     /// Note: If `Err`, the [Wasm] object is no longer guaranteed to be in a valid state
     pub fn read_u8(&mut self) -> Result<u8> {
-        let value = *self.current.first().ok_or(Error::Eof)?;
+        if self.full_contents.len() <= self.pc {
+            return Err(Error::Eof);
+        }
 
-        self.current = self
-            .current
-            .get(1..)
-            .expect("slice to contain at least 1 element");
+        let byte = self.full_contents[self.pc];
 
-        Ok(value)
+        self.pc += 1;
+
+        // if self.full_contents.len() - self.pc == 0 {
+        //     panic!("slice to contain at least 1 element")
+        // }
+
+        Ok(byte)
     }
 
     /// Parses a variable-length `u32` as specified by [LEB128](https://en.wikipedia.org/wiki/LEB128#Unsigned_LEB128).
@@ -66,11 +71,13 @@ impl WasmReader<'_> {
     pub fn read_name(&mut self) -> Result<&str> {
         let len = self.read_var_u32()? as usize;
 
-        if len > self.current.len() {
+        if len > (self.full_contents.len() - self.pc) {
             return Err(Error::Eof);
         }
-        let (utf8_str, rest) = self.current.split_at(len); // Cannot panic because check is done above
-        self.current = rest;
+
+        let (utf8_str, _) = self.full_contents[self.pc..].split_at(len);
+
+        self.pc += utf8_str.len();
 
         core::str::from_utf8(utf8_str).map_err(Error::MalformedUtf8String)
     }
--- a/src/execution/mod.rs
+++ b/src/execution/mod.rs
@@ -3,6 +3,7 @@ use alloc::vec::Vec;
 use value_stack::Stack;
 
 use crate::core::indices::{FuncIdx, GlobalIdx, LocalIdx};
+use crate::core::opcodes::*;
 use crate::core::reader::types::memarg::MemArg;
 use crate::core::reader::types::{FuncType, NumType, ValType};
 use crate::core::reader::{WasmReadable, WasmReader};
@@ -12,7 +13,7 @@ use crate::execution::store::{FuncInst, GlobalInst, MemInst, Store};
 use crate::execution::value::Value;
 use crate::validation::code::read_declared_locals;
 use crate::value::InteropValueList;
-use crate::Error::RuntimeError;
+use crate::Error::{self, RuntimeError};
 use crate::RuntimeError::{DivideBy0, UnrepresentableResult};
 use crate::{Result, ValidationInfo};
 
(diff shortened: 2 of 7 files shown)


src/core/reader/types/values.rs
```diff
@@ -95,6 +95,29 @@ impl WasmReader<'_> {
...
         (0..len).map(|_| read_element(self)).collect()
     }
+
+    pub fn read_instruction(&mut self) -> Result<&[u8]> {
+        if self.current.is_empty() {
+            return Err(Error::Eof);
+        }
+        match self.current {
+            // check if we are at a multibyte (2 byte) instruction
+            [0xFB, _, ..] | [0xFC, _, ..] | [0xFD, _, ..] | [0xFE, _, ..] => {
```
wucke13, line 105: Why don't we use the consts here?

src/core/reader/types/values.rs
wucke13, line 1: Should we maybe remove the `read_instruction` method? At least in the code, with that we are matching two times the bytes then.

src/execution/mod.rs
```diff
@@ -111,42 +112,42 @@ impl<'b> RuntimeInstance<'b> {
         wasm.move_start_to(inst.code_expr);
 
         loop {
-            match wasm.read_u8().unwrap_validated() {
+            match wasm.read_instruction().unwrap_validated() {
```
wucke13, line 115: could this be instead:

```rust
match wasm[self.pc..] {
```

src/execution/mod.rs
```diff
@@ -111,42 +112,42 @@ impl<'b> RuntimeInstance<'b> {
...
                 }
                 // local.set [t] -> []
-                0x21 => {
+                [LOCAL_SET, ..] => {
                     let local_idx = wasm.read_var_u32().unwrap_validated() as LocalIdx;
                     let local = locals.get_mut(local_idx);
                     let value = stack.pop_value(local.to_ty());
                     trace!("Instruction: local.set [{local:?}] -> []");
                     *local = value;
                 }
```
wucke13, line 134: Wherever an instruction is followed by arguments, we could take these directly:

```rust
[LOCAL_SET, local_idx, ..] => {
  let local = locals.get_mut(local_idx);
  let value = stack.pop_value(local.to_ty());
  trace!("Instruction: local.set [{local:?}] -> []");
  *local = value;
}
```

src/validation/code.rs
```diff
@@ -159,44 +160,30 @@ fn read_instructions(
...
-                // Second value
-                assert_pop_value_stack(value_stack, ValType::NumType(NumType::I32))?;
-
+            [I32_CONST, ..] => {
+                let _num = wasm.read_var_i32()?;
                 value_stack.push_back(ValType::NumType(NumType::I32));
             }
-            0x6C => {
-                // First value
+            [I32_ADD, ..] | [I32_MUL, ..] | [I32_DIV_S, ..] | [I32_DIV_U, ..] => {
```
wucke13, line 167: I love that!

src/core/reader/types/values.rs
nerodesu017: You are correct, will fix them to use consts here

src/execution/mod.rs
nerodesu017: I think this could work, we also have to make sure we increment the pc and modify the `current` of the `WasmReader` instance accordingly

src/execution/mod.rs
nerodesu017: good question, will go for this and talk more with @george-cosma that has helped me a lot discussing different ways, hope to get to a solution that is perfect for everyone, both from an aesthetic point of view, but also from a performance point of view

src/core/reader/types/values.rs
nerodesu017: good observation, will discuss with @george-cosma too the perfect way to go about it

src/execution/mod.rs
wucke13: Ideally we could get rid of the current in WasmReader IMHO

src/core/reader/types/values.rs
nerodesu017: Used them in both executor and validator. Please check: 
- [executor](https://github.com/DLR-FT/wasm-interpreter/blob/efa6eac09f56d6ce20aa1ef6efe4198b870522b8/src/execution/mod.rs)
- [validator](https://github.com/DLR-FT/wasm-interpreter/blob/efa6eac09f56d6ce20aa1ef6efe4198b870522b8/src/validation/code.rs)

src/execution/mod.rs
nerodesu017: Done

src/execution/mod.rs
nerodesu017: Moved from `current` being a `&[u8]` to having a `pc` being a `usize`

src/core/reader/types/values.rs
nerodesu017: Removed it and made it look cleaner. Sadly we still have to use the `strip_bytes` command quite often (especially on multibyte instructions - please see [executor/mod.rs](https://github.com/DLR-FT/wasm-interpreter/blob/refactor/opcodes_layout/src/execution/mod.rs), specifically and look at multibyte instructions).

src/execution/mod.rs
nerodesu017: Still an architectural design that I need your input on. For each argument we still need to do a call to `strip_bytes`, do you want me to continue on this path? (read directly from the slice and call `strip_bytes`) Or do you want me to keep the current implementation?

florianhartung: I would just like to add a few arguments against reading the arguments of operations while matching like this: `[SOME_OPCODE, some_argument, ..] => { ... }`.

1. This only works for arguments of fixed size. In WASM, however, integers are encoded with variable length (see [here](https://webassembly.github.io/spec/core/binary/values.html#integers)). Unfortunately, as far as I know, there is no elegant solution for matching them directly in the slice.
2. By matching the arguments directly in the slice, we decouple reading the argument bytes from advancing the program counter, which in my opinion is not very _rusty_. I think it is more error-prone, and puts the additional responsibility on the developer to make sure that the program counter is always incremented by the correct amount.

Don't get me wrong: The `strip_bytes` method is far from the best solution.
Maybe the best way is to write a macro with `macro_rules!` for the whole `match` statement, which allows us to match variable-length arguments. This macro could then also increment the program counter accordingly.

wucke13:
@nerodesu017 Are you ok if I take over this branch, trying to finish the work?

nerodesu017:
> @nerodesu017 Are you ok if I take over this branch, trying to finish the work?

I am fine with that!

wucke13:
I just figured that all the clear changes that we want already landed in 761bd782 . Considering that and the comment in https://github.com/DLR-FT/wasm-interpreter/pull/23#discussion_r1670418377, I would suggest we close this PR. This take on the matching of opcodes (which I unfortunately proposed) has some serious flaws as pointed out in the comment above.

PR #24, opened 2024-07-08
Add deployment for rustdoc and strictdoc on main and preview deployment for requirements

george-cosma:
### Pull Request Overview

This pull request adds a two new github actions:
- `pages_deploy_main` which deploys both the rustdoc and the strictdoc sites to github pages on commit to main.
- `pages_requirement_preview` which deploys only the strictdoc site to github pages on all PRs which modify the requirements. 

### Testing Strategy

This pull request was tested on a [seperate repo](https://github.com/george-cosma/rustdoc-deploy/). For the preview script, two PRs were created - [one which modifies requirements.sdoc](https://github.com/george-cosma/rustdoc-deploy/pull/1) and [one which doesn't](https://github.com/george-cosma/rustdoc-deploy/pull/3).

Result for the example repo [can be found here](https://george-cosma.github.io/rustdoc-deploy/) and the PR1 preview [can be found here](https://george-cosma.github.io/rustdoc-deploy/requirements/pr-preview/pr-1/).

### TODO or Help Wanted

The landing page is extremely barebones. Although not entirely necessary until the MVP, it would still be nice to have something prettier, even if its only job is to contain two links. Should this be addressed in this PR?

### Formatting
N/A

### Github Issue

This pull request helps with #19. Coverage report not included in this PR.

### Author

Signed-off-by: George Cosma [george.cosma@oxidos.io](mailto:george.cosma@oxidos.io)

### DIFF
--- a/.github/workflows/pages_deploy_main.yaml
+++ b/.github/workflows/pages_deploy_main.yaml
@@ -1,0 +1,57 @@
+name: Documentation Deploy
+
+on:
+  # When a PR is merged (or force push to main)
+  push:
+    branches: [ "main" ]
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+# Enable cargo logs to be in colour
+env:
+  CARGO_TERM_COLOR: always
+
+# Sets permissions of the GITHUB_TOKEN to allow deployment to GitHub Pages
+permissions:
+  contents: write
+
+# Allow only one concurrent deployment, skipping runs queued between the run in-progress and latest queued.
+# However, do NOT cancel in-progress runs as we want to allow these production deployments to complete.
+concurrency:
+  group: "pages"
+  cancel-in-progress: false
+
+jobs:
+  deploy:
+    environment:
+      name: github-pages
+      url: ${{ steps.deployment.outputs.page_url }}
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      # -=-=-=-= Rustdocs =-=-=-=-
+      - name: Build rustdocs
+        run: cargo doc
+
+      # -=-=-=-= Strictdoc =-=-=-=-
+      - name: Install python
+        uses: actions/setup-python@v5.1.0
+        with:
+          python-version: 3.12
+      - name: Install strictdoc & requirements
+        run: pip install strictdoc setuptools # Note: we need setuptools for strictdoc to work. Installing just strictdoc is not enough
+      - name: Export requirements
+        run: strictdoc export ./requirements/requirements.sdoc
+
+      # -=-=-=-= Preparation =-=-=-=-
+      - name: Prepare web files
+        run: mv ./target/doc/ ./web/rustdocs/ && mv ./output/html/ ./web/strictdoc/
+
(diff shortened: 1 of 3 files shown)


valexandru:
Looks very nice to me, but with the following comments:

- I don't like the current version of combined documentations. I would either: have two separate independent links for requirements and for the rustdoc (with no common interface) or a way to integrate in the same platform both. I am more inclined to the first option.
- I think the reason why the link is not generated and the pipeline is failing is due to the reason that it does not have access to the repo, which I suspect to be the `read` permission.

george-cosma:
In my last commit I deleted the landing page. I had a an issue with the `main` pipeline. It inherently wants to delete **all** files, then push the changes. This would also delete any existing previews. As such, I made the pipeline only output the compiled html to a subdirectory. This way, if we want a landing page we can set up another action/job to create it, or simply make PRs on the `gh-pages` branch if the content does not require a build step.

.github/workflows/pages_deploy_main.yaml
```diff
@@ -0,0 +1,58 @@
...
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+# Enable cargo logs to be in colour
+env:
+  CARGO_TERM_COLOR: always
+
+# Sets permissions of the GITHUB_TOKEN to allow deployment to GitHub Pages
+permissions:
+  contents: write
```
george-cosma, line 16: TODO: make read @george-cosma

george-cosma:
When the `coverage` label wasn't applied
image

And afterwards:
image

Note: it doesn't fail since it already thinks it deployed a preview (requirement preview). This does mean that we can't preview coverage report and requirements at the same time, but I don't believe that it will ever be an issue.

.github/workflows/pages_deploy_main.yaml
george-cosma: https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs#overview

> For each of the available permissions, shown in the table below, you can assign one of the access levels: read, write, or none. _**`write includes read.`**_ If you specify the access for any of these permissions, all of those that are not specified are set to none

wucke13:
> The landing page is extremely barebones. Although not entirely necessary until the MVP, it would still be nice to have something prettier, even if its only job is to contain two links. Should this be addressed in this PR?

The current state seems function, which is sufficient IMHO. So this can be addressed some time later.

wucke13 (approved):

valexandru (approved):
The only issue that remains is to check the settings on the repository so that the preview is deployed. Maybe @wucke13 you can help us with that

wucke13:
image

I'm not sure what else I need to activate here

george-cosma:
I think that , if merged, this will work. Eventually activate Github Pages with `Deploy from Branch`
image

I'm pretty stumped otherwise too. It works on my repo

george-cosma:
The reason the workflow was failing was because the they started from `george-cosma:actions/docs`. PR #38 makes these changes on this repo, and it successfully deploys. As such, this PR is obsolete.

PR #30, opened 2024-07-11
Enable function invocation by name

george-cosma:
### Pull Request Overview

This pull request adds the ability to invoke a function using its name rather than its id.

This feature is needed as part of parsing `wast` files. Their equivalent of  `assert` [uses the name of functions](https://github.com/WebAssembly/testsuite/blob/83f0863c9f52050590dd47a4647d5c5ef1f2fdf9/i32.wast#L37). Relevant issue: #9 

### Testing Strategy

This pull request was tested by modifying the `add_one` integration test to invoke its function by name rather than by index.

### TODO or Help Wanted

- Change all tests to use named invocation (Q: Do we want to do this, or let the old tests as-is?)
- Q: Should `RuntimeInstance` be responsible for holding onto the exports?

 
### Formatting

- [x] Ran `cargo fmt`
- [x] Ran `cargo check`
- [x] Ran `cargo build`
- [ ] Ran `nix fmt`
- [ ] Ran `treefmt`

### Author

Signed-off-by: George Cosma [george.cosma@oxidos.io](mailto:george.cosma@oxidos.io)


george-cosma:
PR marked as draft until I get some feedback for the question in the "Todo section"

nerodesu017 (approved):
lgtm

valexandru:
Looks good also to me. I see the value in modifying the other tests so that the code is more readable.
Regarding the question, what does it mean holding onto the exports? Storing the references to them? I think so.

florianhartung:
> Change all tests to use named invocation (Q: Do we want to do this, or let the old tests as-is?)

Using named invocations in the tests does not really have any downsides and provides better readability, so I'd say it's worth it.

---
> Should RuntimeInstance be responsible for holding onto the exports?

Yes, the `RuntimeInstance` is our equivalent of the `Module Instance` as specified by the [WASM Spec](https://webassembly.github.io/spec/core/exec/runtime.html#module-instances)

src/execution/mod.rs
```diff
@@ -68,6 +71,29 @@ where
         Ok(instance)
     }
 
+    pub fn invoke_named<Param: InteropValueList, Returns: InteropValueList>(
+        &mut self,
+        func_name: &str,
+        param: Param,
+    ) -> Result<Returns> {
+        let func_idx = self.exports.iter().find_map(|export| {
```
wucke13, line 79: This is a linear search through all registered functions. That is fine for now, but will require optimization later. Please add a `// TODO optimize function search` just for us to pick it up eventually, please.

wucke13 (approved):

valexandru (approved):

wucke13:
The changes lgtm, however I would wish for the commit history to be reworked into two commits, one for the initial draft, one for the change of the tests.

wucke13:
@george-cosma please refactor the two commit messages to be compliant with [conventional commit](https://www.conventionalcommits.org/en/v1.0.0/). If you rebase from main, you will also get a CI check for that.

PR #35, opened 2024-07-17
fix all cargo doc warnings

wucke13:
### Pull Request Overview

This pull request fixes all warnings in the cargo doc departement

### Testing Strategy

`nix flake check`

### Formatting

- [ ] Ran `cargo fmt`
- [ ] Ran `cargo check`
- [ ] Ran `cargo build`
- [x] Ran `nix fmt`
- [ ] Ran `treefmt`

### Author

Signed-off-by: wucke13 wucke13@gmail.com

PR #41, opened 2024-07-17
rework WasmReader

wucke13:
### Pull Request Overview

This pull request rework's the `WasmReader` module.

- Add documentation
- Replace the double slice construct by Slice + Index
- Add tests
- Mark functions const where possible
- Mark `make_span` as unchecked
- Make move_start failable

### Testing Strategy

This pull request was tested by `cargo test`

### TODO or Help Wanted

This pull request requires #35

### Formatting

- [x] Ran `cargo fmt`
- [x] Ran `cargo check`
- [x] Ran `cargo build`
- [x] Ran `nix fmt`
- [x] Ran `treefmt`

### Author

Signed-off-by: wucke13 <wucke13@gmail.com>

### DIFF
--- a/src/core/reader/mod.rs
+++ b/src/core/reader/mod.rs
@@ -4,71 +4,124 @@ use crate::{Error, Result};
 pub mod section_header;
 pub mod types;
 
-/// A struct for managing and reading WASM bytecode.
-/// Its purpose is mostly to abstract parsing basic WASM values from the bytecode.
+/// A struct for managing and reading WASM bytecode
+///
+/// Its purpose is to abstract parsing basic WASM values from the bytecode.
 pub struct WasmReader<'a> {
-    pub(crate) full_contents: &'a [u8],
-    pub(crate) current: &'a [u8],
+    /// Entire WASM binary as slice
+    pub full_contents: &'a [u8],
+
+    /// Current program counter, i. e. index of the next byte to be consumed from the WASM binary
+    pub pc: usize,
 }
 
 impl<'a> WasmReader<'a> {
-    pub fn new(wasm: &'a [u8]) -> Self {
+    /// Initialize a new [WasmReader] from a WASM byte slice
+    pub const fn new(wasm: &'a [u8]) -> Self {
         Self {
             full_contents: wasm,
-            current: wasm,
+            pc: 0,
         }
     }
-    // TODO this is not very intuitive but we cannot shorten `self.current`'s end
-    //  because some methods rely on the property that `self.current`'s and
-    //  `self.full_contents`'s last element are equal.
-    pub fn move_start_to(&mut self, span: Span) {
-        self.current =
-            &self.full_contents[span.from../* normally we would have the end of the span here*/];
-    }
 
-    pub fn remaining_bytes(&self) -> &[u8] {
-        self.current
-    }
-
-    pub fn current_idx(&self) -> usize {
-        self.full_contents.len() - self.current.len()
-    }
-    pub fn make_span(&self, len: usize) -> Span {
-        Span::new(self.current_idx(), len)
-    }
-
-    pub fn strip_bytes<const N: usize>(&mut self) -> Result<[u8; N]> {
-        if N > self.current.len() {
+    /// Advance the cursor to the first byte of the provided [Span] and validates that entire [Span] fits the WASM binary
+    ///
+    /// # Note
+    ///
+    /// This allows one to set the [pc](WasmReader::pc) to one byte *past* the end of
+    /// [full_wasm_binary](WasmReader::full_wasm_binary), **if** the [Span]'s length is 0.
+    pub fn move_start_to(&mut self, span: Span) -> Result<()> {
+        if span.from + span.len > self.full_contents.len() {
             return Err(Error::Eof);
         }
 
-        let (bytes, rest) = self.current.split_at(N);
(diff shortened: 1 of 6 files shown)


src/core/reader/mod.rs
```diff
@@ -4,73 +4,123 @@ use crate::{Error, Result};
...
+            "pc was advanced backwards towards the start"
+        );
+
+        let num_read_bytes = self.pc - before;
         Ok((ret, num_read_bytes))
     }
 
+    /// Skip `num_bytes`, advancing the [pc](Self::pc) accordingly
+    ///
+    /// This can move the [pc](Self::pc) past the last byte of the WASM binary, so that reading
```
florianhartung, line 108: The documentation is a bit contradictory here.
I'd suggest something like this instead:
```
If the resulting [pc](Self::pc) would end up past the last byte of the WASM binary,
an error is returned.
```

florianhartung (changes requested):

src/core/reader/mod.rs
george-cosma: This is also sort-of linked with the commend on `move_start_to`. `pc` can be set to any valid index within the slice of `wasm_bytes`, but also one past that.

|              Index |   0  |   1  |   2  |   3  |   4  | 5 | 6 |
|-------------------:|:----:|:----:|:----:|:----:|:----:|:-:|:-:|
|  wasm_bytes[index] | 0xaa | 0xbb | 0xcc | 0xee | 0xff | - | - |
| Valid `pc` position? |   ✅  |   ✅  |   ✅  |   ✅  |   ✅  | ✅ | ❌ |

The `pc` (if I understood correctly) means "the next position to read from", so, after consuming all valid wasm_bytes, the `pc` will end up at one index past the end. This comment says that the `pc` can **not** be moved by `skip` into an invalid position, but it can leave it in the "end" position, making any future reads panic.

Perhaps the wording on this is a bit vague, and "past the last byte in `wasm_bytes`" is a bit wordy. Perhaps we can explain this phenomenon in the documentation of the struct/field and then reference it in the function comments.

src/core/reader/mod.rs
```diff
@@ -4,73 +4,123 @@ use crate::{Error, Result};
...
 
-    pub fn current_idx(&self) -> usize {
-        self.full_contents.len() - self.current.len()
-    }
-    pub fn make_span(&self, len: usize) -> Span {
-        Span::new(self.current_idx(), len)
+    /// Create a [Span] starting from [pc](Self::pc) for the next `len` bytes
+    ///
+    /// Does **not** verify the span to fit the WASM binary, i.e. the span could exceed the WASM
+    /// binarie's bounds.
```
valexandru, line 51:
```suggestion
    /// binary's bounds.
```

src/core/reader/mod.rs
```diff
@@ -84,17 +134,25 @@ pub mod span {
 
     use crate::core::reader::WasmReader;
 
+    /// A index and offset to describe a (sub-) slice into WASM bytecode
```
valexandru, line 137:
```suggestion
    /// An index and offset to describe a (sub-) slice into WASM bytecode
```

src/core/reader/mod.rs
```diff
@@ -84,17 +134,25 @@ pub mod span {
...
 
     impl Span {
-        pub fn new(from: usize, len: usize) -> Self {
+        /// Create a new [Span], starting from `from` and ranging `len` elements
+        pub const fn new(from: usize, len: usize) -> Self {
             Self { from, len }
         }
-        pub fn len(&self) -> usize {
+
+        /// Returnt the length of this [Span]
```
valexandru, line 154:
```suggestion
        /// Returns the length of this [Span]
```

src/core/reader/mod.rs
```diff
@@ -103,7 +161,81 @@ pub mod span {
...
+
+    #[test]
+    fn move_start_to() {
+        let my_bytes = vec![0x11, 0x12, 0x13, 0x14, 0x15];
+        let mut wasm_reader = WasmReader::new(&my_bytes);
+
+        let span = Span::new(0, 0);
+        wasm_reader.move_start_to(span).unwrap();
+        // this actually dangerous, we did not validate there to be more than 0 bytes using the Span
+        wasm_reader.peek_u8().unwrap();
```
valexandru, line 182: Yep, it is a bit weird that the Span is used *actually* only for moving the pc forward with a number of bytes. Even though the Span also contains a length.

florianhartung: FYI: The idea of having a `Span` object was just an idea of mine when I implemented the first code expression parsing.
It was used to mark where a code block starts and ends, but I think we might be able to refactor its API to be a little less obscure. Maybe we could also get rid of it completely🤷‍♂️

wucke13: For now we can keep the span's, they have some value in validation IIRC, because they allow to specify exactly the span of bytes for a given WASM item.

src/core/reader/mod.rs
wucke13: I added a lot of documentation and more tests, and I copied your Table @valexandru . I hope this is now better documented.

valexandru (approved):

george-cosma (approved):

florianhartung (approved):

Issue #42, opened 2024-07-17
How to design the API for resumable function calls


wucke13:
In order to satisfy the ["Resumable" requirement](https://github.com/DLR-FT/wasm-interpreter/blob/132c2e5d624ac4d787af83d53c42679aba88ab87/requirements/requirements.sdoc#L7) we need to define an API that allows for a halted function call to continue.

There are several ways of achieving this, which we can discuss in the comments.

wucke13:
# Low hanging fruit: expose stack & pc, have `Interpreter::run()` return when out of fuel

We let the user manipulate the stack themselves to push arguments/pop return values. Calling a function is merely setting the program counter to the function's first instruction and running the interpreter. The interpreter can return at any time, irrespective of whether the WASM function finished. Successful completion of the function is signaled by a boolean returned form the `run` method of the interpreter.

- Allows for abitrary WASM function signatures, that are only known at run-time
- Provides no compile-time type safety for calling WASM code
- Allows messy things, like calling one function, then after the interpreter returned (but before the function completed) calling just another function
- Is likely a good candidate for a low-level, possibly only internal API

# Mock-up Code

```rust
let interpreter = RuntimeInstance { /* ... */ };

interpreter.stack.push_i32(0x13); // push first arg
interpreter.stack.push_i32(0x42); // second arg
interpreter.wasm_reader.pc = interpreter.get_fn_index("add").unwrap();
interpreter.add_fuel(10_000);
let mut completed = interpreter.run();
/* react to out of fuel, maybe notify someone */
while !completed {
  interpreter.add_fuel(150);
  completed = interpreter.run();
  /* react to out of fuel, maybe notify someone */
}
let return_value = interpreter.stack.pop();
```

florianhartung:
This is a solution that @wucke13 and I have been working on.
It works similar to the previous code example, but provides a bit more safety.



```rs
let interpreter: RuntimeInstance = todo!("create runtime instance");

// `FunctionDefinition` contains the function signature and the WASM entry point
let add_function: FunctionDefinition = interpreter.get_fn_index("add").expect("a function named `add` to exists");

// Let's create a new Runner. It starts off in the state `RunnerState::Initializing`.
let mut runner = Runner::new(&mut interpreter);

// Put some values on the stack
runner.push_stack(&[
  WasmValue::I32(0x13),
  WasmValue::I32(0x42),
]).expect("stack to be modifiable after runner initialisation");

// We can add fuel at any time
[...]

PR #46, opened 2024-07-19
Dynamic types for function invokation

### LINKED ISSUE #43: Enable calling WASM function who's type signature is only known at run-time

george-cosma:
### Pull Request Overview

This pull request adds function invocation with parameter & return type only known at runtime.

### Testing Strategy

This pull request was tested by only running `cargo test`.

### TODO or Help Wanted

I need help to stabilize the API for this. A notable missing feature is dynamic invocation by name. As it is, I would need to double the amount of functions. This isn't a big issue, the code repetition isn't that bad, but I want to hear your thoughts.

### Formatting

- [x] Ran `cargo fmt`
- [x] Ran `cargo check`
- [x] Ran `cargo build`
- [x] Ran `cargo doc`
- [ ] Ran `nix fmt`
- [ ] Ran `treefmt`

### Github Issue

This pull request closes #43

### DIFF
--- a/src/execution/mod.rs
+++ b/src/execution/mod.rs
@@ -24,7 +24,7 @@ pub mod hooks;
 pub(crate) mod label;
 pub(crate) mod locals;
 pub(crate) mod store;
-pub(crate) mod value;
+pub mod value;
 pub mod value_stack;
 
 pub struct RuntimeInstance<'b, H = EmptyHookSet>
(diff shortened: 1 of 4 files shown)


george-cosma:
The reason `invoke_dynamic_unchecked`  was to avoid a little bit of code repetition. I made `invoke_dynamic` by modifying `invoke_func`, however I noticed that the part where the stack & parameters are initialized, and the call to start interpreting are shared between the two.

I initially wanted to abstract this away into another function and `invoke_dynamic_unchecked` was born. Now `invoke_func` checks parameter & return type that are known compile-time and `invoke_dynamic` does the same, but with types that are only known at runtime. Both then call into `invoke_dynamic_unchecked`  to start the interpeter. A more apt name for `invoke_dynamic_unchecked`  is `run_func`.

The `unsafe` was added just to mark the fact that no checks are made in `invoke_dynamic_unchecked`, and the end-user needs to do it themself. `invoke_func` and `invoke_dynamic` do these checks.

Whilst writing this comment, I started to question if it is necessary, but I guess that the interface will change anyways when #42 is tackled.

nerodesu017 (approved):
I guess it'd be okay to have the two commits squashed into one? Either way, lgtm

wucke13 (approved):

PR #54, opened 2024-07-25
 Convert panics to errors

george-cosma:
### Pull Request Overview

This pull request attempts to convert most panics into errors.

### Testing Strategy

This pull request was tested by running `cargo test`

### TODO or Help Wanted

There are some `.expects()` that I'm unsure HOW to turn to errors, or if we should. I'm also a bit afraid that our `Error` type is becoming "too precise" or "over-specific" (i.e. An error variant is ever created only once). I'd like some input on this.

### Formatting

- [x] Ran `cargo fmt`
- [x] Ran `cargo check`
- [x] Ran `cargo build`
- [x] Ran `cargo doc`
- [ ] Ran `nix fmt`
- [ ] Ran `treefmt`

### DIFF
--- a/src/core/error.rs
+++ b/src/core/error.rs
@@ -1,4 +1,4 @@
-use crate::core::indices::GlobalIdx;
+use crate::core::indices::{FuncIdx, GlobalIdx};
 use core::fmt::{Display, Formatter};
 use core::str::Utf8Error;
 
@@ -38,8 +38,12 @@ pub enum Error {
     InvalidMutType(u8),
     MoreThanOneMemory,
     InvalidGlobalIdx(GlobalIdx),
+    InvalidFunctionIdx(FuncIdx),
     GlobalIsConst,
     RuntimeError(RuntimeError),
+    NotYetImplemented(&'static str),
+    InvalidParameterTypes,
+    InvalidResultTypes,
 }
 
 impl Display for Error {
@@ -108,12 +112,24 @@ impl Display for Error {
             Error::InvalidGlobalIdx(idx) => f.write_fmt(format_args!(
                 "An invalid global index `{idx}` was specified"
             )),
+            Error::InvalidFunctionIdx(idx) => f.write_fmt(format_args!(
+                "An invalid function index `{idx}` was specified"
+            )),
             Error::GlobalIsConst => f.write_str("A const global cannot be written to"),
             Error::RuntimeError(err) => match err {
                 RuntimeError::DivideBy0 => f.write_str("Divide by zero is not permitted"),
                 RuntimeError::UnrepresentableResult => f.write_str("Result is unrepresentable"),
                 RuntimeError::FunctionNotFound => f.write_str("Function not found"),
             },
+            Error::NotYetImplemented(msg) => {
+                f.write_fmt(format_args!("Not yet implemented: {msg}"))
+            }
+            Error::InvalidParameterTypes => {
+                f.write_str("The given parameter types do not match the function's signature")
+            }
+            Error::InvalidResultTypes => {
+                f.write_str("The given result types do not match the function's signature")
+            }
         }
     }
 }
--- a/src/execution/mod.rs
+++ b/src/execution/mod.rs
@@ -14,7 +14,7 @@ use crate::execution::store::{FuncInst, GlobalInst, MemInst, Store};
 use crate::execution::value::Value;
 use crate::validation::code::read_declared_locals;
 use crate::value::InteropValueList;
-use crate::Error::RuntimeError;
+use crate::Error::{self, RuntimeError};
 use crate::RuntimeError::{DivideBy0, FunctionNotFound, UnrepresentableResult};
 use crate::{Result, ValidationInfo};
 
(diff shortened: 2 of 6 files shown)


src/core/error.rs
```diff
@@ -38,8 +38,12 @@ pub enum Error {
     InvalidMutType(u8),
     MoreThanOneMemory,
     InvalidGlobalIdx(GlobalIdx),
+    InvalidFunctionIdx(FuncIdx),
     GlobalIsConst,
     RuntimeError(RuntimeError),
+    NotYetImplemented(&'static str),
```
wucke13, line 44: I'm opposing to introduce this, is there any reason to not just use `todo!()`?

src/execution/mod.rs
```diff
@@ -169,7 +179,12 @@ where
         let error = self.function(func_idx, &mut stack);
         error?;
 
-        let func_inst = self.store.funcs.get(func_idx).expect("valid FuncIdx");
+        // TODO(george-cosma): Do we want this to be a RuntimeError(FunctionNotFound)?
```
wucke13, line 182: My vote is yes!

src/execution/mod.rs
```diff
@@ -263,9 +278,9 @@ where
                                 let address = address as usize;
                                 mem.data.get(address..(address + 4))
                             })
-                            .expect("TODO trap here");
+                            .expect("TODO trap here"); // ???
```
wucke13, line 281: Should this be an out-of-bounds access to linear memory?

wucke13 (approved):
In general, really good! I'm struggling whether I like having "not yet implemented" as an error variant.

Further on, we might need its own trap error enum, that explicitly handles all reasons for a trap of the interpreter.

src/core/error.rs
george-cosma: While sketching out a runner for the wasm specsuite, I ran into the `todo!`s for the sections a lot of times. I wanted to have avoid a `catch_unwind` call more due to esthetic reasons.

florianhartung: I'm also against introducing this error variant. IMO it's okay to have `catch_unwind` in the test runner. Maybe we could even check if the panic message includes "not yet implemented" and display a simple statistic.

florianhartung:
> One general note on this: I think we should `.expect("something")` whenever the validation ensures that we are always on the Ok/Some side of things. It would be misleading to introduced `?` on places where it is statically known that the None/Err case never occurs. This might ease MC/DC testing later, as it at least clearly discerns the branches that we believe are actually impossible.

@wucke13 Just a quick note.. That's what we have this for: https://github.com/DLR-FT/wasm-interpreter/blob/6f91fc5dab7597de13ae486a6a468850646bd0b7/src/execution/assert_validated.rs#L5-L19

cemonem:
Is there a reason we are trying to specify why the traps arise? There is no mention of specific trap codes in the spec, all traps are encoded with the _trap_ instruction https://webassembly.github.io/spec/core/exec/runtime.html#administrative-instructions

> The _trap_ instruction represents the occurrence of a trap. Traps are bubbled up through nested instruction sequences, ultimately reducing the entire program to a single instruction, signalling abrupt termination.

Maybe we can simply get away without implementing an error for everything that causes a trap and simply have an error for trap. Otherwise there might always be things we miss, or traps that correspond to more than one error.

cemonem:
or maybe there might be benefit, other wasm interpreters seem to be doing this. Copying the codes off of them and iterating would be the best I guess

george-cosma:
For me, it's not super clear what a trap ought to be in terms of core wasm. My interpretation of the spec is that a trap is something similar to rust's panic, though I don't know for sure.

Perhaps it would be correct to say that our compiler can exit due to 3 reasons at runtime:
1. Invalid runtime state (tried to run and indirect_call instruction to a NULL function reference) - in this case that is an issue with the code, we return an appropriate error and we stop all execution
2. Invalid wasm code state / "panic" in wasm code. Let's say that someone compiled some wasm code to represent a game of tic tac toe and the turn order variable goes to player 3 (or something else equally invalid). This could be caught by an `unwrap`, `index out of bounds`, or just a plain `if  turn > 2 {panic!("AAAaa")}`. In this case, our interpreter would hit the "trap" instruction and we would send out a generic error for "trap".
3. Illegal interpreter state - we have a fatal bug, or we hit a `todo!()`

>or maybe there might be benefit, other wasm interpreters seem to be doing this. Copying the codes off of them and iterating would be the best I guess

@cemonem Could you link some examples, I'm curious now 😄. Also @nerodesu017 with the instant `👍` reaction 😆

nerodesu017:
While we could get away with simply having a general Error that resembles a Trap (see https://github.com/WebAssembly/wabt/blob/2e23b86f2716877f6e24df0c39ac8f0f0e64c770/src/interp/interp.cc#L1238 for example), it helps with code readability, maintainability and testing (especially this part, I love me the intellisense when asserting for traps)

Besides we have more granular control (for example seeing if an error was at runtime or not, etc)

cemonem:
> @cemonem Could you link some examples, I'm curious now 😄. Also @nerodesu017 with the instant `👍` reaction 😆
The latest memory implementation from @nerodesu017 had a link: https://github.com/wasmi-labs/wasmi/blob/37d1449524a322817c55026eb21eb97dd693b9ce/crates/core/src/trap.rs#L220

george-cosma:
Yhea, imo I think we can do it how wasmi is doing it

PR #60, opened 2024-07-30
fix: rework github workflows to use nix

george-cosma:
### Pull Request Overview

This pull request changes the github workflows to use nix. The coverage preview was left unmodified since I believe it works well-enough as-is.

### Testing Strategy

This pull request was tested on my own fork. See https://github.com/george-cosma/wasm-interpreter/pull/9 for coverage-preview and https://github.com/george-cosma/wasm-interpreter/actions/runs/10161441119/job/28099916237 for main deployment

### Formatting

- [ ] Ran `cargo fmt`
- [ ] Ran `cargo check`
- [ ] Ran `cargo build`
- [ ] Ran `cargo doc`
- [x] Ran `nix fmt`
- [ ] Ran `treefmt`

### DIFF
--- a/.github/workflows/pages_deploy_main.yaml
+++ b/.github/workflows/pages_deploy_main.yaml
@@ -22,57 +22,28 @@ concurrency:
   cancel-in-progress: false
 
 jobs:
-  deploy:
+  build-reports:
+    name: Build Reports using Nix
     environment:
       name: github-pages
       url: ${{ steps.deployment.outputs.page_url }}
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
-      # -=-=-=-= Rustdocs =-=-=-=-
-      - name: Build rustdocs
-        run: cargo doc
-
-      # -=-=-=-= Strictdoc =-=-=-=-
-      - name: Install python
-        uses: actions/setup-python@v5.1.0
+      - uses: cachix/install-nix-action@v27
         with:
-          python-version: 3.12
-      - name: Install strictdoc & requirements
-        run: pip install strictdoc setuptools # Note: we need setuptools for strictdoc to work. Installing just strictdoc is not enough
-      - name: Export requirements
-        run: strictdoc export ./requirements/requirements.sdoc
-
-      # -=-=-=-= Preparation (1/2) =-=-=-=-
-      - name: Prepare web files (1/2)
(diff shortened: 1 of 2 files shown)


.github/workflows/pages_deploy_main.yaml
```diff
@@ -22,57 +22,28 @@ concurrency:
...
+          authToken: ${{ secrets.CACHIX_AUTH_TOKEN }}
+      - run: nix build .#wasm-interpreter --print-build-logs
+      - run: nix build .#report --print-build-logs
       - name: Deploy to Github Pages
         uses: peaceiris/actions-gh-pages@v4.0.0
         with:
           github_token: ${{ secrets.GITHUB_TOKEN }}
           # Build output to publish to the `gh-pages` branch:
-          publish_dir: ./web
+          publish_dir: ./result
```
wucke13, line 47: As `./result` is a symlink, at least some programs refuse to work with it, while `./result/` is already the resolved symlink. I hope this is not an issue here, but just a note, sometimes `./result` doesn't work but `./result/` does, because the one is just a symlink, while the other is already the resolved target of that symlink.

wucke13:
Note: There is one caveat, which justifies still running cargo test external from Nix. Nix either succeeds or fails. If it fails, we get no report at all. So in order to have some insight on test failure having a _normal_ cargo test CI run that is not in `nix build` still is very valuable. For these cases, `nix develop . --command cargo test` could be useful, which just uses Nix to provide the toolchain but then runs arbitrary commands in shell (without the all-or-nothin logic of `nix build`).

george-cosma:
I'm not sure I am seeing the issue? For the main deployment, we know that it will (most likely) because it has gone through a PR review process. In the case of the coverage report preview, if the build/tests fail, I don't why a coverage report can't be generated? I'm not sure I understand why you would want an extra task to run `cargo test`

.github/workflows/pages_deploy_main.yaml
george-cosma: I've had plenty of issues with the result folder but not because it is a symlink but because the output folder is owned by root. `peaceiris/actions-gh-pages` doesn't seem to have any issues with it from my testing

wucke13 (approved):

florianhartung (approved):
LGTM (i have no clue what this is)

PR #65, opened 2024-08-06
feat: f64 operations

nerodesu017:
### Pull Request Overview

This pull request adds f64-related op codes.

### Testing Strategy

This pull request was tested locally.

### TODO or Help Wanted

N/A

### Formatting

- [x] Ran `cargo fmt`
- [x] Ran `cargo check`
- [x] Ran `cargo build`
- [x] Ran `cargo doc`
- [ ] Ran `nix fmt`
- [ ] Ran `treefmt`

### Github Issue

This pull request implements issue #2


src/execution/mod.rs
```diff
@@ -24,6 +24,7 @@ pub(crate) mod locals;
 pub(crate) mod store;
 pub mod value;
 pub mod value_stack;
+// pub mod ;
```
valexandru, line 27:
```suggestion
```

src/core/reader/types/values.rs
```diff
@@ -40,6 +40,22 @@ impl WasmReader<'_> {
...
+            self.full_wasm_binary[self.pc..self.pc + 8]
+                .try_into()
+                .unwrap(),
+        );
+
+        self.strip_bytes::<8>()?;
+
+        Ok(word)
+    }
+
```
florianhartung, line 58: Would this be possible here instead?

```suggestion
        let bytes = self.strip_bytes::<8>().map_err(|_| Error::Eof)?;
        let word = u64::from_le_bytes(bytes);
```

florianhartung (changes requested):

src/core/reader/types/values.rs
florianhartung: This should also be possible:
```rs
let bytes = self.strip_bytes::<{ std::mem::size_of::<u64>() }>().map_err(|_| Error::Eof)?;
```

nerodesu017: Used first suggestion

florianhartung (approved):

george-cosma (approved):

valexandru (approved):

Issue #39, opened 2024-07-17
Make function recursion possible


wucke13:
The code in 

https://github.com/DLR-FT/wasm-interpreter/blob/67d4df2c36c5b7fb6aa0fda78af3c665e94e654b/src/execution/mod.rs#L110

does not allow for nested/recursive function calls while adhering to the ["Stackless" requirement](https://github.com/DLR-FT/wasm-interpreter/blob/67d4df2c36c5b7fb6aa0fda78af3c665e94e654b/requirements/requirements.sdoc#L54). In particular, we need to store each stack frame's locals in the `RuntimeInstance`.

PR #68, opened 2024-08-15
Dev/wucke13 refactor 3

### LINKED ISSUE #63: Implement stack frames for functions on the `Stack` struct
The changes from https://github.com/DLR-FT/wasm-interpreter/pull/47 actually contained a misconception; they re-implemented the `Stack` struct partially. We need instead to fix the `Stack` struct implementation, so that it both contains the value- and the call-frame stack.

wucke13:
### Pull Request Overview

This pull request adds recursive function calls

### Testing Strategy

This pull request was tested `cargo test`

### TODO or Help Wanted

This pull request still needs additional vetting

### Formatting

- [ ] Ran `cargo fmt`
- [ ] Ran `cargo check`
- [ ] Ran `cargo build`
- [ ] Ran `cargo doc`
- [ ] Ran `nix fmt`
- [ ] Ran `treefmt`

### Github Issue

This pull request closes #63, #39


src/validation/code.rs
```diff
@@ -96,6 +107,35 @@ fn read_instructions(
...
+
+                let ret_vals_start = value_stack.len() - this_func_ty.returns.valtypes.len();
+                let _remaining_locals = value_stack.drain(..ret_vals_start);
+
+                // TODO(george-cosma): There is a bug here. Because there is another op code after this reteurn (an
+                // `end`), I have to let this function consume it to work for the tests.
+            }
+            // call [t1*] -> [t2*]
+            CALL => {
+                // TODO get function params, reverse iter over it, for each param pop that value from the value_stack
```
george-cosma, line 125: I think we can remove this TODO, right?

src/validation/code.rs
```diff
@@ -96,6 +107,35 @@ fn read_instructions(
...
+
+                if value_stack.len() < this_func_ty.returns.valtypes.len() {
+                    return Err(Error::EndInvalidValueStack); // TODO(george-cosma): Is this the right error?
+                }
+
+                let ret_vals_start = value_stack.len() - this_func_ty.returns.valtypes.len();
+                let _remaining_locals = value_stack.drain(..ret_vals_start);
+
+                // TODO(george-cosma): There is a bug here. Because there is another op code after this reteurn (an
+                // `end`), I have to let this function consume it to work for the tests.
```
george-cosma, line 121: As discussed previously, this is likely not a bug, since there is no evidence two `end`s  will ever be emitted by a compiler. 

The issue comes when a hand-made `wat` file which contains two `end`s is passed to the interpreter. In this case, after finishing with validation I propose we check that we reached the end of the function (all bytes were consumed).

src/validation/code.rs
```diff
@@ -96,6 +107,35 @@ fn read_instructions(
             END => {
                 return Ok(());
             }
+            RETURN => {
+                let this_func_ty = &fn_types[type_idx_of_fn[this_function_idx]];
+
+                if value_stack.len() < this_func_ty.returns.valtypes.len() {
+                    return Err(Error::EndInvalidValueStack); // TODO(george-cosma): Is this the right error?
```
george-cosma, line 114: Needs to be resolved

src/execution/value_stack.rs
```diff
@@ -12,81 +13,147 @@ use crate::unreachable_validated;
...
 }
+
+/// The [WASM spec](https://webassembly.github.io/spec/core/exec/runtime.html#stack) calls this `Activations`, however it refers to the call frames of functions.
+pub(crate) struct CallFrame {
+    /// Index to the function of this CallFrame
+    pub func_idx: FuncIdx,
+
+    /// Local varaiables such as parameters for this [CallFrame]'s function
+    /// TODO(george-cosma): Do we want "locals" to be a Box<\[Value\]>? Do we want two layers of indirection when getting locals?
+    pub locals: Locals,
```
george-cosma, line 152: This comments should be resolved

src/execution/value_stack.rs
```diff
@@ -12,81 +13,147 @@ use crate::unreachable_validated;
...
+    // TODO figure out if this has to change the value stack
+    pub fn push_stackframe(&mut self, func_idx: FuncIdx, locals: Locals, return_addr: usize) {
+        // TODO maybe manipulate the value stack
+        self.frames.push(CallFrame {
+            func_idx,
+            locals,
+            return_addr,
+            stack_values: 0,
+        })
+    }
```
george-cosma, line 122: I'm not sure I understand these TODOs

george-cosma (changes requested):

src/validation/code.rs
wucke13: LGTM!

src/execution/value_stack.rs
wucke13: I think we can leave it as is. Currently `Locals` is a one-field struct over a `Box<[Value]>`, so I would expect the compiler to not actually introduced a second level of indirection.

george-cosma: I've noticed that we rely a lot on the compiler doing smart optimizations, yet we never check for it explicitly. Perhaps we should keep track of our assumptions, and eventually verify them (if we can?)

wucke13: That definitely is a good though. Though there are so many micro decisions that I would like refactor anyhow once we have the base set of features, that I'm not sure if it makes sense to spent the effort on tracking our assumptions.

One way of doing this could be writing unit tests that just assert the size of types, i.e. in this case the size of `Locals` to be that of an `usize`, or more precisely of a `Box<[ValTypes]>`.

george-cosma: I agree with you that it is early. I suggest we start investigating this matter further after we have stabilized our architecture

src/execution/mod.rs
```diff
@@ -118,7 +118,10 @@ where
             params.into_values().into_iter(),
             func_inst.locals.iter().cloned(),
         );
-        stack.push_stackframe(func_idx, locals, 0);
+
+        // setting `usize::MAX` as return address for the outermost function ensures that we
+        // observably fail upon errornoeusly contining execution after that function returns.
```
george-cosma, line 123: typo: `contining`

george-cosma (dismissed):

wucke13:
With the git wrangling I rewrote all commits that @george-cosma contributed to. I added them as co-author however, to correctly reflect that these commits contain significant contribution from them.

I hope, that this is now finally ready for merge.

src/execution/interpreter_loop.rs
```diff
@@ -68,26 +60,68 @@ pub(super) fn run<H: HookSet>(
 
         match first_instr_byte {
             END => {
-                break;
+                let maybe_return_address = stack.pop_stackframe();
+
+                // We finished this entire invocation if there is no tackframe left. If there are
```
valexandru, line 65:
```suggestion
                // We finished this entire invocation if there is no stackframe left. If there are
```

src/execution/value_stack.rs
```diff
@@ -12,81 +13,174 @@ use crate::unreachable_validated;
...
-                    unreachable_validated!()
-                }
-            }
-            StackEntry::Label(_) => todo!("pop label from stack"),
-            StackEntry::Activation(_) => todo!("pop activation frame from stack"),
+        // If there is at least one stack frame, we shall not pop values past the current
+        // stackframe. However, there is one legitimate reason to pop when there is **no** current
+        // stackframe: after the outermost function returns, to extract the final return values of
+        // this interpreter invocation.
+        debug_assert!(
```
valexandru, line 38: A question here, maybe we discussed about this, but would we want an error state/panic/etc. in the release mode if we manage to get here?

wucke13: Yes this is only there to find bugs when running `cargo test`. Basically, assuming that there are no logic bugs in the implementation, this will never panic.

wucke13: It did show me an intersting logic bug though: I inserted a `debug_assert` to verify that values are never popped from the value stack past the current stack frames base pointer. Turns out, there is a moment where there is 0 `CallFrame`s, but is still legal to pop values from the stack: after an invocation finished, to retrieve the return values.

src/execution/interpreter_loop.rs
wucke13: Good catch, fixed!

florianhartung (approved):

george-cosma (approved):

cemonem:
implemented in https://github.com/DLR-FT/wasm-interpreter/blob/400533d972c06ab8d16afee759417cb40c7ca5c7/src/execution/interpreter_loop.rs#L114

PR #69, opened 2024-08-15
feat(core): transactions for the wasm reader

florianhartung:
### Pull Request Overview

This pull request adds a transaction wrapper function for the WASM reader.

### Testing Strategy

This pull request was not tested.

### Formatting

- [X] Ran `cargo fmt`
- [x] Ran `cargo check`
- [X] Ran `cargo build`
- [x] Ran `cargo doc`
- [X] Ran `nix fmt`
- [x] Ran `treefmt`

### Github Issue

This pull request does not close any issues.


florianhartung:
The `map_err` call can be replaced with the newer `inspect_err`, which would be a bit cleaner IMO.
However using [`Result::inspect_err`](https://doc.rust-lang.org/std/result/enum.Result.html#method.inspect_err) would change our MSRV to 1.76.0, so I was unsure if I could use it.

wucke13 (dismissed):

src/core/reader/mod.rs
```diff
@@ -148,9 +149,28 @@ impl<'a> WasmReader<'a> {
...
 }
 
 pub trait WasmReadable: Sized {
+    /// Reads a new [`Self`] from given [`WasmReader`].
+    ///
+    /// Note that if this function returns `Err(_)`, the [`WasmReader`] may still have been advanced,
+    /// which may lead to unexpected behaviour.
+    /// To avoid this consider using the [`WasmReader::handle_transaction`] method to wrap this function call.
     fn read(wasm: &mut WasmReader) -> Result<Self>;
     fn read_unvalidated(wasm: &mut WasmReader) -> Self;
```
valexandru, line 175: When would the unvalidated function be used?

valexandru (approved):

src/core/reader/mod.rs
wucke13: I added documentation to that function in the last commit :smile:

wucke13 (approved):

george-cosma (approved):
LGTM

PR #78, opened 2024-08-27
Feat/globals

george-cosma:
### Pull Request Overview

This pull request adds module-level support for globals. 
Note: not all constant expressions are implemented. The missing ones are:
- `ref.null`
- `ref.func`
- `global.get` (need to properly implement imports first)

### Testing Strategy

This pull request was tested by running `cargo test` + a lot of manually testing in wat2wasm. Details of my findings are in the comments

### Help Wanted

I feel like the code is everywhere. A little advice as to how to organize all the new additions would be welcome!

### Formatting

- [x] Ran `cargo fmt`
- [x] Ran `cargo check`
- [x] Ran `cargo build`
- [ ] Ran `cargo doc`
- [ ] Ran `nix fmt`

### Github Issue

This pr depends on #77


src/validation/code.rs
```diff
@@ -357,7 +357,120 @@ fn read_instructions(
...
+            return Err(Error::ExprMissingEnd);
+        };
+        trace!("Read cosntant instruction byte {first_instr_byte:#X?} ({first_instr_byte})");
+
+        // Valid constant instructions are:
+        // - Core: https://webassembly.github.io/spec/core/valid/instructions.html#valid-constant
+        // - Extended Proposal: https://webassembly.github.io/extended-const/core/valid/instructions.html#valid-constant
+        use crate::core::reader::types::opcode::*;
+        match first_instr_byte {
+            // Missing: ref.null, ref.func, global.get
```
wucke13, line 389: These extensive comments are really nice, can we make them doc string of the surrounding function?

src/execution/mod.rs
```diff
@@ -241,8 +241,13 @@ where
...
             .map(|global| {
-                // TODO execute `global.init_expr` to get initial value. For now just use a default value.
-                let value = Value::default_from_ty(global.ty.ty);
+                let mut wasm = WasmReader::new(validation_info.wasm);
+                // The place we are moving the start to should, by all means, be inside the wasm bytecode.
+                wasm.move_start_to(global.init_expr).unwrap_validated();
+                let mut stack = Stack::new();
+
+                run_const(wasm, &mut stack, ());
+                let value = stack.pop_value(global.ty.ty);
```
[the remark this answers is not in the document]
george-cosma: I am not sure if it the spec allows for it but both wasmtime and wat2wasm do not allow for garbage to be left behind. Here's an example code if you want to play around with it yourself.

```wat
(module
  (global (export "g") (mut i32)
    i32.const 5
    i32.const 5
    i32.const 5
    i32.const 5
    i32.const 5
    i32.const 10
    i32.add
  )
  (func (export "f")
    i32.const 100
    global.set 0))
```

george-cosma:
As a note, besides what we discussed I also moved `validate_value_stack` to `validation/mod.rs`

src/core/reader/types/global.rs
```diff
@@ -1,28 +1,47 @@
...
+            Ok(())
+        })?;
+
+        // At this point, we can assume that `init_expr` is `Some(_)`. `read_constant_instructions` returns a Span or an
+        // Error. If an Error is returned it is pushed up the call stack.
         Ok(Self {
-            ty: GlobalType::read(wasm)?,
-            init_expr: (), /* todo!("read constant expression") */
+            ty,
+            init_expr: init_expr.unwrap_validated(),
```
florianhartung, line 39: Is there a way around unwrapping here?
Also I think this should be a normal `.unwrap()` or `.expect()` call, because for it to be `None` the `validate_value_stack` method would have to return `Ok(())` without calling the closure.

florianhartung (changes requested):
I'm not sure whether it is the right approach to have duplicated validations and implementations for all const expressions (see `execution::const_interpreter_loop::run_const`).

george-cosma:
> I'm not sure whether it is the right approach to have duplicated validations and implementations for all const expressions (see `execution::const_interpreter_loop::run_const`).

While I agree that having both another validator and an execution loop for const expressions is a bit inelegant - tons of repeated code! - the alternative would be to integrate these expressions into the regular loops. While possible, this would take a lot of work and I'd wager that the end result would be less readable than what we have now. Just to name a few reasons, the const expression loop needs to return the length of its code (it isn't mentioned in the binary, like how it is for code blocks); a merged execution loop would still need to differentiate between regular execution and const execution, because we want to error out if we try to execute a non-const instruction in a const expression; etc.

src/core/reader/types/global.rs
george-cosma: I could create a default-value span to avoid it. The issue lies in the fact that `validate_value_stack` returns an empty result. I could technically modify it to be able to return a generic type. Or I could just replace the `unwrap_validated` with an `expect`. Your call.

florianhartung: Hmm... Using a generic for the function might be misleading. Let's stick with an expect 👍

florianhartung (dismissed):

nerodesu017 (approved):

src/execution/const_interpreter_loop.rs
```diff
@@ -0,0 +1,101 @@
...
+            }
+            I64_MUL => {
+                let v1: i64 = stack.pop_value(ValType::NumType(NumType::I64)).into();
+                let v2: i64 = stack.pop_value(ValType::NumType(NumType::I64)).into();
+                let res = v1.wrapping_mul(v2);
+
+                trace!("Constant instruction: i64.mul [{v1} {v2}] -> [{res}]");
+                stack.push_value(res.into());
+            }
+            _ => panic!("¯\\_(ツ)_/¯"),
```
florianhartung, line 98: Please add a proper "invalid instruction" message. You can keep the lil guy here

florianhartung (changes requested):

src/execution/const_interpreter_loop.rs
george-cosma: In the main interpreter loop, we put out a trace message and just skip the instruction. Should we do that here too, instead of panic?

florianhartung: Hmm.. I think that panicking both here and in the interpreter_loop might be better to catch errors ASAP.
Please leave this as a panic for now. I will create an issue for the interpreter_loop.

george-cosma: Okay, modified it in both places. What do you think?
In regards the the little buddy, I've talked to him and convinced him to free up the spot. Knowing him, he will come back, but not in this PR.

florianhartung (approved):

wucke13 (approved):

PR #87, opened 2024-09-19
feat: Linker

### LINKED ISSUE #83: The never-ending Linker problem
# The Linker Problem

To fully implement a working wasm interpreter we must be able to resolve imports. The idea of a `Linker` comes to mind. There are multiple ways to design it.

## Design 1: Monolithic Runtime Instance 
 
linker_problem_monolith drawio

This design would entail collecting all validation info from all of the modules into the Linker, which will then produce a "merged" validation info which can then be instantiated as a sole RuntimeInstance. Import resolution would then be resolved universally, since a call to an imported function would actually be a regular call for the merged validation info.

**✅ Pros:**
- Import Resolution solved at link-time. No performance overhead.
- The architecture of the code will remain mostly unchanged
- Linker will live for a very short amount of time (long enough to process the validation infos)

**❌Cons:**
- Code sections will (likely) need to be modified due to index namespace merger. A call to function "12" is trivial with a single module, but what about if we merge two modules? Module 1 imports a function as index 1, and Module 2 exports this function with index 8. All function indices will then have to be reshuffled and the appropriate `call` instructions will need their operands changed.
- Same as above, but for globals, tables, memories, etc.
- Import Resolution solved at link-time. This would prevent any sort of design where equivalent modules could be swapped at runtime.

## Design 2: Each validation info has its own runtime instance (Swarm)

linker_problem_swarm drawio

As the name and image suggests, each module gets its own runtime instance. Where the magic lies is actually inside the Linker, which, this time, is an entity which lives as long as the runtimes. When a module needs to call an imported function, it does so via the Linker.

**✅ Pros:**
- Validation Info & Runtime creation is simple
- Closer in style to the WASM Component Model (TODO: Check specification)
- Allows for the ability to "hot swap" modules at runtime. Let's say the infotainment system is decoupled (intentionally or not). Then the Linker could be ordered to replace the module responsible for communicating with it, with a dummy module that fails instantly instead of waiting for a timeout.

**❌Cons:**
- Performance overhead due to import resolution at runtime, especially if we tackle the threading proposal.
- The fuel mechanic is non-trivial
- Lots of arhitectural changes
- Linker is long-running
- A very specific problem with resumability (continue reading)

With this type of linker, there is an arhitectural problem we need to solve to maintain resumability. If Module 1 calls an imported function in Module 2, and Module 2 then calls an imported function from Module 1, at the end of this chain Module 1 must be able to resume code properly.

Here is as example how an import call could work:

linker_problem_rel1 drawio

Now, how it would work in the described scenario:

linker_problem_rel2 drawio

Notice that the second `Store PC` would overwrite the previously stored program counter. An intuitive solution would be to make it a stack, but that ***feels*** like it would create more problems than it solves. An alternative solution would have the `call` instruction create a callframe not only on the caller module, but also on the called module. Or something like that. There are solutions, but I do not know which is the correct one.

I'd like to continue this discussion. I want to know your opinions in regards with which approach to go with. I personally believe the second "swarm" approach is more appropriate, but that is based more on *vibes*.

george-cosma:
### Pull Request Overview

This pull request adds a "Lookup table" linker. See #83 first.

Basically, after each time a new module is added we attempt to create a "lookup table". This lookup table translate the imported functions from each module to the exported local function in the already-existing modules.

Resumability is kept by passing all module-specific details (like the wasm binary, the readers, function types, store, etc.) to the interpreter_loop::run function. These details have been compiled into a structure called `ExecutionInfo`.

Also of note, I split the "FunctionInstance" structure into two variants - local and imported, for clarity sake. This comes with the needed modification to the validation process and store-creation process.

### Testing Strategy

This pull request was tested by writing a new test, and tested against the existing tests.

### Formatting

- [ ] Ran `cargo fmt`
- [ ] Ran `cargo check`
- [ ] Ran `cargo build`
- [ ] Ran `cargo doc`
- [ ] Ran `nix fmt`

### Github Issue

This pull request closes #83

### DIFF
--- a/src/execution/execution_info.rs
+++ b/src/execution/execution_info.rs
@@ -1,0 +1,29 @@
+use alloc::string::{String, ToString};
+use alloc::vec::Vec;
+
+use crate::core::reader::types::FuncType;
+use crate::core::reader::WasmReader;
+use crate::execution::Store;
+
+/// ExecutionInfo is a compilation of relevant information needed by the [interpreter loop](
+/// crate::execution::interpreter_loop::run). The lifetime annotation `'r` represents that this structure needs to be
+/// valid at least as long as the [RuntimeInstance](crate::execution::RuntimeInstance) that creates it.
+pub struct ExecutionInfo<'r> {
+    pub name: String,
+    pub wasm_bytecode: &'r [u8],
+    pub wasm_reader: WasmReader<'r>,
+    pub fn_types: Vec<FuncType>,
+    pub store: Store,
+}
+
+impl<'r> ExecutionInfo<'r> {
+    pub fn new(name: &str, wasm_bytecode: &'r [u8], fn_types: Vec<FuncType>, store: Store) -> Self {
+        ExecutionInfo {
+            name: name.to_string(),
+            wasm_bytecode,
+            wasm_reader: WasmReader::new(wasm_bytecode),
+            fn_types,
+            store,
+        }
+    }
+}
--- a/src/execution/interpreter_loop.rs
+++ b/src/execution/interpreter_loop.rs
@@ -44,7 +44,9 @@ pub(super) fn run<H: HookSet>(
     let func_inst = store
         .funcs
         .get(stack.current_stackframe().func_idx)
-        .unwrap_validated();
+        .unwrap_validated()
+        .try_into_local()
+        .expect("to start execution with local function");
 
     // Start reading the function's instructions
     let mut wasm = WasmReader::new(wasm_bytecode);
@@ -78,6 +80,8 @@ pub(super) fn run<H: HookSet>(
                     .funcs
                     .get(stack.current_stackframe().func_idx)
                     .unwrap_validated()
+                    .try_into_local()
+                    .expect("TODO: interpreter_loop")
                     .code_expr;
                 if wasm.pc != current_func_span.from() + current_func_span.len() {
                     continue;
@@ -100,6 +104,8 @@ pub(super) fn run<H: HookSet>(
                     .funcs
                     .get(stack.current_stackframe().func_idx)
                     .unwrap_validated()
+                    .try_into_local()
+                    .expect("TODO: interpreter_loop")
                     .sidetable;
             }
             IF => {
@@ -160,7 +166,12 @@ pub(super) fn run<H: HookSet>(
             CALL => {
                 let func_to_call_idx = wasm.read_var_u32().unwrap_validated() as FuncIdx;
 
-                let func_to_call_inst = store.funcs.get(func_to_call_idx).unwrap_validated();
+                let func_to_call_inst = store
+                    .funcs
+                    .get(func_to_call_idx)
+                    .unwrap_validated()
+                    .try_into_local()
+                    .expect("TODO: interpreter_loop");
                 let func_to_call_ty = types.get(func_to_call_inst.ty).unwrap_validated();
 
                 let params = stack.pop_tail_iter(func_to_call_ty.params.valtypes.len());
@@ -205,7 +216,9 @@ pub(super) fn run<H: HookSet>(
                 let func_to_call_inst = store
                     .funcs
                     .get(func_addr.unwrap_validated())
-                    .unwrap_validated();
+                    .unwrap_validated()
+                    .try_into_local()
+                    .expect("TODO: interpreter_loop");
 
                 let func_ty_actual_index = func_to_call_inst.ty;
 
--- a/src/execution/interpreter_loop.rs
+++ b/src/execution/interpreter_loop.rs
@@ -18,13 +18,13 @@ use crate::{
     core::{
         indices::{DataIdx, FuncIdx, GlobalIdx, LabelIdx, LocalIdx, TableIdx, TypeIdx},
         reader::{
-            types::{memarg::MemArg, BlockType, FuncType},
+            types::{memarg::MemArg, BlockType},
             WasmReadable, WasmReader,
         },
         sidetable::Sidetable,
     },
     locals::Locals,
-    store::{DataInst, Store},
+    store::{DataInst, FuncInst},
     value::{self, FuncAddr, Ref},
     value_stack::Stack,
     Limits, NumType, RefType, RuntimeError, ValType, Value,
@@ -33,23 +33,26 @@ use crate::{
 #[cfg(feature = "hooks")]
 use crate::execution::hooks::HookSet;
 
+use super::{execution_info::ExecutionInfo, lut::Lut};
+
 /// Interprets a functions. Parameters and return values are passed on the stack.
 pub(super) fn run<H: HookSet>(
-    wasm_bytecode: &[u8],
-    types: &[FuncType],
-    store: &mut Store,
+    modules: &mut [ExecutionInfo],
+    current_module_idx: &mut usize,
+    lut: &Lut,
     stack: &mut Stack,
     mut hooks: H,
 ) -> Result<(), RuntimeError> {
-    let func_inst = store
+    let func_inst = modules[*current_module_idx]
+        .store
         .funcs
         .get(stack.current_stackframe().func_idx)
         .unwrap_validated()
         .try_into_local()
-        .expect("to start execution with local function");
+        .unwrap_validated();
 
     // Start reading the function's instructions
-    let mut wasm = WasmReader::new(wasm_bytecode);
+    let mut wasm = &mut modules[*current_module_idx].wasm_reader;
 
     // the sidetable and stp for this function, stp will reset to 0 every call
     // since function instances have their own sidetable.
@@ -63,7 +66,7 @@ pub(super) fn run<H: HookSet>(
     loop {
         // call the instruction hook
         #[cfg(feature = "hooks")]
-        hooks.instruction_hook(wasm_bytecode, wasm.pc);
+        hooks.instruction_hook(modules[*current_module_idx].wasm_bytecode, wasm.pc);
 
         let first_instr_byte = wasm.read_u8().unwrap_validated();
 
@@ -76,18 +79,23 @@ pub(super) fn run<H: HookSet>(
                 // just skip because it is a delimiter of a ctrl block
 
                 // TODO there is definitely a better to write this
-                let current_func_span = store
+                let current_func_span = modules[*current_module_idx]
+                    .store
                     .funcs
                     .get(stack.current_stackframe().func_idx)
                     .unwrap_validated()
                     .try_into_local()
-                    .expect("TODO: interpreter_loop")
+                    .unwrap_validated()
                     .code_expr;
+
+                // There might be multiple ENDs in a single function. We want to
+                // exit only when the outermost block (aka function block) ends.
                 if wasm.pc != current_func_span.from() + current_func_span.len() {
                     continue;
                 }
 
-                let (maybe_return_address, maybe_return_stp) = stack.pop_stackframe();
+                let (return_module, maybe_return_address, maybe_return_stp) =
+                    stack.pop_stackframe();
 
                 // We finished this entire invocation if there is no stackframe left. If there are
                 // one or more stack frames, we need to continue from where the callee was called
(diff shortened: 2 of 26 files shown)


src/execution/interpreter_loop.rs
```diff
@@ -33,19 +35,22 @@ use crate::execution::hooks::HookSet;
 
 /// Interprets a functions. Parameters and return values are passed on the stack.
 pub(super) fn run<H: HookSet>(
-    wasm_bytecode: &[u8],
-    types: &[FuncType],
-    store: &mut Store,
+    modules: &mut [ExecutionInfo],
+    current_module: &mut usize,
```
nerodesu017, line 39: Should we rename `current_module` here to `current_module_idx` and take `modules[*current_module]` outside of each match arm and put it before the loop so it's cleaner?
What do you guys think?

src/execution/value_stack.rs
```diff
@@ -106,8 +106,9 @@ impl Stack {
     }
 
     /// Pop a [`CallFrame`] from the call stack, returning the return address
```
nerodesu017, line 108:
```suggestion
    /// Pop a [`CallFrame`] from the call stack, returning the module id and return address
```

george-cosma:
Blcoked on #88

src/execution/mod.rs
```diff
@@ -277,7 +323,7 @@ where
             })
             .ok_or(RuntimeError::FunctionNotFound)?;
 
-        Ok((0, func_idx))
+        Ok((module_idx, func_idx))
     }
 
     fn verify_function_ref(
```
george-cosma, line 329: Document this +, in the first if, ModuleNotFound

src/execution/mod.rs
```diff
@@ -259,10 +299,16 @@ where
     // TODO: replace this with the lookup table when implmenting the linker
     fn get_indicies(
```
george-cosma, line 300: Remove todo, document

src/execution/mod.rs
```diff
@@ -259,10 +299,16 @@ where
...
         &self,
-        _module_name: &str,
+        module_name: &str,
         function_name: &str,
     ) -> Result<(usize, usize), RuntimeError> {
-        let func_idx = self
+        let module_idx = *self
+            .module_map
+            .get(module_name)
+            .ok_or(RuntimeError::FunctionNotFound)?;
```
george-cosma, line 308: Module not found

src/execution/mod.rs
```diff
@@ -151,18 +166,36 @@ where
...
+            .try_into_local()
+            .ok_or(RuntimeError::FunctionNotFound)?;
+        let func_ty = self.modules[module_idx]
+            .fn_types
+            .get(func_inst.ty)
+            .unwrap_validated();
 
         // Check correct function parameters and return types
         if func_ty.params.valtypes != Param::TYS {
-            panic!("Invalid `Param` generics");
```
george-cosma, line 162: Revert to panic!("Invalid `Param` generics");

src/execution/mod.rs
```diff
@@ -151,18 +166,36 @@ where
...
+            panic!(
+                "Invalid `Param` generics. Expected: {:?}, Found: {:?}",
+                func_ty.params.valtypes,
+                Param::TYS
+            );
         }
         if func_ty.returns.valtypes != Returns::TYS {
-            panic!("Invalid `Returns` generics");
```
george-cosma, line 165: Revert to panic!("Invalid `Returns` generics");

src/execution/mod.rs
```diff
@@ -227,20 +270,17 @@ where
...
-        stack.push_stackframe(func_idx, func_ty, locals, 0);
+        stack.push_stackframe(module_idx, func_idx, &func_ty, locals, 0);
 
         // Run the interpreter
         run(
-            self.wasm_bytecode,
-            &self.types,
-            &mut self.store,
+            &mut self.modules,
+            &mut 0, // todo!
```
george-cosma, line 278: todo!

src/execution/mod.rs
```diff
@@ -227,20 +270,17 @@ where
...
+            self.lut.as_ref().ok_or(RuntimeError::DivideBy0)?,
```
george-cosma, line 279: UnmetImportsError

src/execution/mod.rs
```diff
@@ -297,11 +343,19 @@ where
         } else {
             let (module_idx, func_idx) = (function_ref.module_index, function_ref.function_index);
 
-            // TODO: verify module named - index mapping.
+            let module = self
+                .modules
+                .get(module_idx)
+                .ok_or(RuntimeError::FunctionNotFound)?;
```
george-cosma, line 349: Module not found

src/execution/mod.rs
```diff
@@ -297,11 +343,19 @@ where
...
+
+            if module.name != function_ref.module_name {
+                return Err(RuntimeError::FunctionNotFound);
```
george-cosma, line 352: Module not found

src/execution/mod.rs
```diff
@@ -288,20 +357,30 @@ where
             let (module_idx, func_idx) =
                 self.get_indicies(&function_ref.module_name, &function_ref.function_name)?;
 
-            if module_idx != function_ref.module_index || func_idx != function_ref.function_index {
-                // TODO: should we return a different error here?
+            if module_idx != function_ref.module_index {
```
nerodesu017, line 360: Maybe just one `if` statement here with both conditions combined?

src/execution/mod.rs
```diff
@@ -67,16 +73,15 @@ where
...
-            exports: validation_info.exports.clone(),
-            store,
+            modules: Vec::new(),
+            module_map: BTreeMap::new(),
+            lut: None,
             hook_set,
         };
+        instance.add_module(module_name, validation_info);
 
+        // TODO: how do we handle the start function, if we don't have a LUT yet?
```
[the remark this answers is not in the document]
nerodesu017: Running an example where we have an import and it's not met in [wasm-interp](https://github.com/WebAssembly/wabt) leads to the same error: invalid import. I think this is the safe way to go about it.
(wasm-interp actually errors out even if we don't run any of the code:

```wat
(module
  (import "console" "log" (func $log (param i32)))
  (func $dummy
    i32.const 0
    drop 
  )
;;   (start $dummy)
)
```
)

src/execution/mod.rs
george-cosma: The reason I wanted to split them off was to be able to discern which reference is wrong - the module or the function. This was done to conform to the two errors: ModuleNotFound and FunctionNotFound. However, it would also be valid to say there should be a 3rd error type to indicate a stale or incorrect reference

nerodesu017: I see, that makes sense. I also agree on the 3rd error type for an incorrect reference.

src/execution/lut.rs
```diff
@@ -0,0 +1,113 @@
...
+            })
+            .find_map(|desc| {
+                if let ExportDesc::FuncIdx(func_idx) = desc {
+                    Some((*module_idx, *func_idx))
+                } else {
+                    None
+                }
+            })
+    }
+}
```
cemonem, line 113: This scheme can't handle cases where imports cascade. Let A, B, C be modules where A imports func B.f and rexports A.g, and C import A.g as C.h. When h is called, we need to recursively follow until a local func is hit (which we might not). we might need to check the func_idx is not imported itself here, and if it is, follow the other module without a recursive func call (stackless interpreter rule). We should also throw error if the (module_idx,func_idx) ends up the same as local_module_idx, local_func_idx, since that indicates a cyclic reference. But we can also make do with this too and leave it as an issue.

tests/imports.rs
```diff
@@ -0,0 +1,82 @@
...
+#[test_log::test]
+pub fn run_simple_import() {
+    let wasm_bytes = wat::parse_str(SIMPLE_IMPORT_BASE).unwrap();
+    let validation_info = validate(&wasm_bytes).expect("validation failed");
+    let mut instance =
+        RuntimeInstance::new_named("base", &validation_info).expect("instantiation failed");
+
+    let wasm_bytes = wat::parse_str(SIMPLE_IMPORT_ADDON).unwrap();
+    let validation_info = validate(&wasm_bytes).expect("validation failed");
+    instance.add_module("env", &validation_info);
```
cemonem, line 74: This a bit nitpicky, but I am more in favor of an interface where if A depends on B's store values for instantiation and B depends on nothing, B is instantiated with a RuntimeInstance::new_named("B") and then we instantiate a with RuntimeInstance::new_named_with("A", instance_of_B/imports??) sort of thing, since instantiation of A might depend on already instantiated const values etc. in B (https://webassembly.github.io/spec/core/exec/modules.html#external-typing) here I think the spec necessitates that the imports are instantiated within their module store. For globals at least this is definitely required, since a global.get of an imported const global can be used in constant exprs during instantiation, and that itself needs to have been instantiated. I think in js module instantiation is also this way, but i am not sure. But this can stay too.

cemonem (changes requested):

wucke13 (approved):

cemonem (approved):

src/execution/mod.rs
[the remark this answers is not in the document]
george-cosma: Actually that's not entirely true. While yes, it does not work because it expects the import to be resolved BEFORE running start, if the import is resolvable we can use start on any function, even an imported one. The tricky thing is, I'm not 100% sure if the spec allows it. I will convert this to an issue
https://webassembly.github.io/spec/core/valid/modules.html?highlight=start#xref-syntax-modules-syntax-start-mathsf-func-x

src/execution/mod.rs
george-cosma: TODO, I'm not too bothered now. We will rework this in the future

src/validation/mod.rs
```diff
@@ -73,10 +73,30 @@ pub fn validate(wasm: &[u8]) -> Result<ValidationInfo> {
...
+        })?
+        .unwrap_or_default();
+
+    let imported_functions = imports
+        .iter()
+        .filter_map(|import| match &import.desc {
+            ImportDesc::Func(type_idx) => Some(*type_idx),
+            _ => None,
+        })
+        .collect::<Vec<_>>();
```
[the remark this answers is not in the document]
george-cosma: Fixed in next commit

src/execution/execution_info.rs
```diff
@@ -0,0 +1,29 @@
+use alloc::string::{String, ToString};
+use alloc::vec::Vec;
+
+use crate::core::reader::types::FuncType;
+use crate::core::reader::WasmReader;
+use crate::execution::Store;
+
+/// ExecutionInfo is a compilation of relevant information needed by the [interpreter loop](
```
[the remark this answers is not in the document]
george-cosma: /shrug, this will be modified as we find we need more and more stuff into the interpreter_loop function. This isn't even meant to be public. I agree, but I'll fix this after we start mass rework/documentation :D

tests/imports.rs
george-cosma: This gives more freedom in the linking process to the user. While it may not follow the logic of dependency chains initially, when we attempt to truly interact with the runtime that logic is then enforced. A future dependency graph may be required, but we will have to see...

src/execution/lut.rs
george-cosma: While true, that honestly seems like a skill issue on whoever made this dependency injection. Plus we don't even currently allow to go from an imported function to another import. the LUT doesn't accept that. See `LUT::manual_lookup`. While it might be intresting to tackle, let's have someone yell at us first and we'll get around to "fixing" it

george-cosma: And  also, even if it loops, if between function call some global state is modified, then we do not know if its cyclical. And import to import to import is not allowed by this arhitecture since import to import is not allowed. I'll add a test to enforce this

src/execution/mod.rs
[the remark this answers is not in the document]
wucke13: I interprete the spec as:

- using an imported function as start is ok
- the start function must have the functype `[] -> []`
- the start function of all modules can only be ran once the modules are otherwise fully initialized, e.g const expressions have executed and imports are fully resolved

PR #88, opened 2024-09-23
Linker API Changes

### LINKED ISSUE #83: The never-ending Linker problem

george-cosma:
### Pull Request Overview

This pull request changes the RuntimeInstance API to be compatible with #87 linker proposal. See details #83.

### Formatting

- [x] Ran `cargo fmt`
- [x] Ran `cargo check`
- [x] Ran `cargo build`
- [ ] Ran `cargo doc`
- [ ] Ran `nix fmt`

### Github Issue

This pull request helps close #83


george-cosma:
I recommend looking at the first commit and the last. The one in the middle only changes the tests.

src/execution/mod.rs
```diff
@@ -65,43 +78,81 @@ where
...
+            .iter()
+            .find(|export| match &export.desc {
+                ExportDesc::FuncIdx(idx) => *idx == function_idx,
+                _ => false,
+            })
+            .map(|export| export.name.clone())
+            .ok_or(RuntimeError::FunctionNotFound)?;
+
+        Ok(FunctionRef {
+            module_name: DEFAULT_MODULE.to_string(),
```
nerodesu017, line 130: add TODO comment

src/execution/function_ref.rs
```diff
@@ -0,0 +1,35 @@
...
+        }
+    }
+
+    pub fn invoke<H: HookSet, Param: InteropValueList, Returns: InteropValueList>(
+        &self,
+        runtime: &mut RuntimeInstance<H>,
+        params: Param,
+    ) -> Result<Returns, RuntimeError> {
+        runtime.invoke(self, params)
+    }
```
florianhartung, line 27: I don't think we need this method. The user should call `runtime.invoke(...)` directly (the tests already do it like this).

src/execution/mod.rs
```diff
@@ -65,43 +78,82 @@ where
...
-            if export.name == func_name {
-                match export.desc {
-                    ExportDesc::FuncIdx(idx) => Some(idx),
-                    _ => None,
-                }
-            } else {
-                None
-            }
-        });
+    pub fn get_fn(
```
florianhartung, line 97: Please rename this and `get_fn_idx` to `get_function_by_name` and `get_function_by_index`.
Especially `get_fn_idx` is misleading as the name implies that is should return an index.

src/execution/mod.rs
```diff
@@ -65,43 +78,82 @@ where
...
+            // Note: function name is not important here, as it is not used in the verification process.
+            let start_fn = FunctionRef {
+                module_name: module_name.to_string(),
+                function_name: "start".to_string(),
+                module_index: 0,
+                function_index: start,
+                exported: false,
+            };
+            let result = instance.invoke::<(), ()>(&start_fn, ());
             result?;
```
florianhartung, line 91:
```suggestion
            let _result = instance.invoke::<(), ()>(&start_fn, ())?;
```

benches/hook_performance_impact.rs
```diff
@@ -52,14 +52,16 @@ fn criterion_benchmark(c: &mut Criterion) {
     }
 
     let mut instance_non_empty_hookset =
-        RuntimeInstance::new_with_hooks(&validation_info, MyCustomHookSet).unwrap();
+        RuntimeInstance::new_with_hooks(DEFAULT_MODULE, &validation_info, MyCustomHookSet).unwrap();
 
     c.bench_function("invoke_func EmptyHookSet", |b| {
-        b.iter(|| instance_empty_hookset.invoke_func::<_, ()>(black_box(2), black_box(42_i32)))
+        let test_fn = instance_empty_hookset.get_fn_idx(0, 2).unwrap();
```
florianhartung, line 58: I think it is more common to do the setup for benchmarks outside of `bench_function`.
Criterion.rs has only example that kinda explains this for benchmarks with inputs:
```rs
// Setup (construct data, allocate memory, etc)
let input = 5u64;
c.bench_with_input(
    BenchmarkId::new("function_name", input), &input,
    |b, i| b.iter(|| {
        // Code to benchmark using input `i` goes here
    }),
);
```
https://docs.rs/criterion/latest/criterion/struct.Criterion.html#example-1

florianhartung (changes requested):
LGTM

benches/hook_performance_impact.rs
george-cosma: I put the `test_fn` initialization outside the `bench_function`. I believe `bench_with_input` is more appropriate for something with an input that can vary? I'm not sure, though.

src/execution/function_ref.rs
george-cosma: I'm indifferent. I added it to see what other people think about it. We'll discuss this at the meeting and go on from there.

florianhartung (approved):

nerodesu017 (approved):
LGTM

PR #90, opened 2024-09-24
fix: make pages_deploy_main.yaml to pull submodules

george-cosma:
Github pages dceployment fails now. Woops.
Bug introduced in #32


nerodesu017 (approved):
Can't actually test without pushing so... Let's try?

george-cosma:
Works on my branch: https://github.com/george-cosma/wasm-interpreter/deployments/github-pages

florianhartung (approved):

PR #99, opened 2024-11-25
feat: elements

nerodesu017:
### Pull Request Overview

This pull request adds support for elements and tables.

### Testing Strategy

This pull request was tested using the provided tests which are ported from the [official testsuite](https://github.com/WebAssembly/testsuite).

### TODO or Help Wanted

This pull request still needs your help so that I can make it production-ready.

### Formatting

- [x] Ran `cargo fmt`
- [x] Ran `cargo check`
- [x] Ran `cargo build`
- [x] Ran `cargo doc`
- [ ] Ran `nix fmt`

### Github Issue

This pull request partially implements #2

### DIFF
--- a/src/core/reader/types/element.rs
+++ b/src/core/reader/types/element.rs
@@ -1,7 +1,8 @@
 use super::RefType;
 use crate::core::reader::span::Span;
 use crate::core::reader::WasmReader;
-use crate::read_constant_expression::read_constant_instructions;
+use crate::read_constant_expression::read_constant_expression;
+use crate::validation_stack::ValidationStack;
 use crate::{Error, Result};
 
 use alloc::collections::btree_set::BTreeSet;
@@ -83,7 +84,12 @@ impl ElemType {
                     return Err(Error::UnknownTable);
                 }
 
-                let init_expr = read_constant_instructions(wasm, None, None, Some(functions))?;
+                let mut valid_stack = ValidationStack::new();
+                let init_expr =
+                    read_constant_expression(wasm, &mut valid_stack, None, None, Some(functions))?;
+
+                // on top of the stack it's supposed to be the
+                valid_stack.assert_pop_val_type(super::ValType::NumType(super::NumType::I32))?;
 
                 ElemMode::Active(ActiveElem {
                     table_idx,
@@ -114,7 +120,39 @@ impl ElemType {
             let items: ElemItems = if third_bit_set {
                 ElemItems::Exprs(
                     reftype_or_elemkind.unwrap_or(RefType::FuncRef),
-                    wasm.read_vec(|w| read_constant_instructions(w, None, None, Some(functions)))?,
+                    wasm.read_vec(|w| {
+                        let mut valid_stack = ValidationStack::new();
+                        let span = read_constant_expression(
+                            w,
+                            &mut valid_stack,
+                            None,
+                            None,
+                            Some(functions),
+                        );
+
+                        use crate::validation_stack::ValidationStackEntry::*;
+
+                        if let Some(val) = valid_stack.peek_stack() {
+                            if let Val(val) = val {
+                                match val {
+                                    crate::ValType::RefType(_) => {}
+                                    crate::ValType::NumType(crate::NumType::I32) => {}
+                                    crate::ValType::NumType(_) => {
+                                        return Err(Error::InvalidValidationStackValType(Some(val)))
+                                    }
+                                    _ => {
+                                        return Err(Error::InvalidValidationStackValType(Some(val)))
+                                    }
+                                }
+                            } else {
+                                return Err(Error::InvalidValidationStackType(val));
+                            }
+                        } else {
+                            return Err(Error::InvalidValidationStackValType(None));
+                        }
+
+                        span
+                    })?,
                 )
             } else {
                 assert!(reftype_or_elemkind.is_none());
--- a/src/core/reader/types/element.rs
+++ b/src/core/reader/types/element.rs
@@ -1,0 +1,164 @@
+use super::RefType;
+use crate::core::reader::span::Span;
+use crate::core::reader::WasmReader;
+use crate::read_constant_expression::read_constant_instructions;
+use crate::{Error, Result};
+
+use alloc::collections::btree_set::BTreeSet;
+use alloc::vec::Vec;
+use core::fmt::Debug;
+
+#[derive(Clone)]
+pub struct ElemType {
+    pub init: ElemItems,
+    pub mode: ElemMode,
+}
+
+impl Debug for ElemType {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        write!(
+            f,
+            "ElemType {{\n\tinit: {:?},\n\tmode: {:?},\n\t#ty: {:?}\n}}",
+            self.init,
+            self.mode,
+            self.init.ty()
+        )
+    }
+}
+
+impl ElemType {
+    pub fn ty(&self) -> RefType {
+        self.init.ty()
+    }
+
+    pub fn to_ref_type(&self) -> RefType {
+        match self.init {
+            ElemItems::Exprs(rref, _) => rref,
+            ElemItems::RefFuncs(_) => RefType::FuncRef,
+        }
+    }
+
+    // TODO: @nerodesu017 maybe split up the validation from the parsing?
+    /// Here we can't implement WasmReadable because we also want a mutable
+    /// reference to a [`BTreeSet<u32>`] (`referenced_functions`)
+    ///
+    /// This comes in handy later on when we are validating the actual code of
+    /// the functions so that we can make sure we are not referencing invalid
+    /// functions
+    pub fn read_from_wasm(
+        wasm: &mut WasmReader,
+        functions: &[usize],
+        referenced_functions: &mut BTreeSet<u32>,
+        tables_length: usize,
+    ) -> Result<Vec<Self>> {
+        wasm.read_vec(|wasm| {
+            let prop = wasm.read_var_u32()?;
+
+            // TODO: @nerodesu017 revisit this comment
+            // https://webassembly.github.io/spec/core/syntax/modules.html#element-segments
+            // https://webassembly.github.io/spec/core/binary/modules.html#element-section
+            // We can treat the ty as a 3bit integer
+            // If it's not 3 bits I am not sure what to do
+            // bit 0 => diff between passive|declartive and active segment
+            // bit 1 => presence of an explicit table index for an active segment
+            // bit 2 => use of element type and element expressions instead of element kind and element indices
+            // decide if we should
+
+            // TODO: @nerodesu017 error, this is a parse error, not validation FYI
+            // NOTE: This assert breaks my rustfmt :(
+            // assert!(prop <= 0b111, "Element section is not encoded correctly. The type of this element is over 7 (0b111)");
+
+            let elem_mode = if prop & 0b011 == 0b011 {
+                ElemMode::Declarative
+            } else if prop & 0b001 == 0b001 {
+                ElemMode::Passive
+            } else {
+                let table_idx = if prop & 0b010 == 0b010 {
+                    wasm.read_var_u32()?
+                } else {
+                    0
+                };
+
+                if tables_length <= table_idx as usize {
+                    return Err(Error::UnknownTable);
+                }
+
+                let init_expr = read_constant_instructions(wasm, None, None, Some(functions))?;
+
+                ElemMode::Active(ActiveElem {
+                    table_idx,
+                    init_expr,
+                })
+            };
+
+            let third_bit_set = prop & 0b100 == 0b100;
+
+            let type_kind = if prop & 0b011 != 0 {
+                if third_bit_set {
+                    Some(wasm.read_u8()?)
+                } else {
+                    match wasm.read_u8()? {
+                        0x00 => None,
+                        _ => return Err(Error::OnlyFuncRefIsAllowed),
+                    }
+                }
+            } else {
+                None
+            };
+
+            let reftype_or_elemkind: Option<RefType> = match type_kind {
+                Some(ty) => Some(RefType::from_byte(ty)?),
+                None => None,
+            };
+
+            let items: ElemItems = if third_bit_set {
+                ElemItems::Exprs(
+                    reftype_or_elemkind.unwrap_or(RefType::FuncRef),
+                    wasm.read_vec(|w| read_constant_instructions(w, None, None, Some(functions)))?,
+                )
+            } else {
+                assert!(reftype_or_elemkind.is_none());
+                ElemItems::RefFuncs(wasm.read_vec(|w| {
+                    let offset = w.read_var_u32()?;
+                    referenced_functions.insert(offset);
+                    Ok(offset)
+                })?)
+            };
+
+            let el = ElemType {
+                init: items,
+                mode: elem_mode,
+            };
+
+            Ok(el)
+        })
+    }
+}
+
+#[derive(Debug, Clone)]
+pub enum ElemItems {
+    RefFuncs(Vec<u32>),
+    Exprs(RefType, Vec<Span>),
+}
+
+impl ElemItems {
+    pub fn ty(&self) -> RefType {
+        match self {
+            Self::RefFuncs(_) => RefType::FuncRef,
+            Self::Exprs(rty, _) => *rty,
+        }
+    }
+}
+
+#[derive(Debug, Clone)]
+pub enum ElemMode {
+    Passive,
+    Active(ActiveElem),
+    Declarative,
+}
+
+#[derive(Debug, Clone)]
+pub struct ActiveElem {
+    pub table_idx: u32,
+    pub init_expr: Span,
+}
--- a/src/core/reader/types/mod.rs
+++ b/src/core/reader/types/mod.rs
@@ -7,10 +7,12 @@ use core::fmt::{Debug, Formatter};
 
 use crate::core::reader::{WasmReadable, WasmReader};
 use crate::execution::assert_validated::UnwrapValidatedExt;
+use crate::value::{ExternAddr, FuncAddr, Ref};
 use crate::Result;
 use crate::{unreachable_validated, Error};
 
 pub mod data;
+pub mod element;
 pub mod export;
 pub mod function_code_header;
 pub mod global;
@@ -85,6 +87,28 @@ pub enum RefType {
     ExternRef,
 }
 
+impl RefType {
+    /// TODO: we have to make sure they are NOT null Refs, but still, they are
+    /// not valid ones as we cast them from RefTypes which don't hold addresses
+    /// per-se
+    pub fn to_null_ref(&self) -> Ref {
+        match self {
+            RefType::ExternRef => Ref::Extern(ExternAddr::null()),
+            RefType::FuncRef => Ref::Func(FuncAddr::null()),
+        }
+    }
+}
+
+impl RefType {
+    pub fn from_byte(byte: u8) -> Result<RefType> {
+        match byte {
+            0x70 => Ok(RefType::FuncRef),
+            0x6F => Ok(RefType::ExternRef),
+            _ => Err(Error::InvalidRefType),
+        }
+    }
+}
+
 impl WasmReadable for RefType {
     fn read(wasm: &mut WasmReader) -> Result<RefType> {
         let ty = match wasm.peek_u8()? {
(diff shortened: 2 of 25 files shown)


src/validation/read_constant_expression.rs
```diff
@@ -131,7 +132,22 @@ pub fn read_constant_instructions(
                 todo!();
```
nerodesu017, line 132: I'll make sure to include the `REF_NULL` updated, after gathering all feedback

src/core/reader/types/element.rs
```diff
@@ -0,0 +1,165 @@
...
+        self.init.ty()
+    }
+    pub fn to_ref_type(&self) -> RefType {
+        match self.init {
+            ElemItems::Exprs(rref, _) => rref,
+            ElemItems::RefFuncs(_) => RefType::FuncRef,
+        }
+    }
+
+    pub fn read_from_wasm(
```
nerodesu017, line 40: We couldn't just blindly implement `WasmReadable` for ElemType, as we need to pass multiple arguments to the reading function

src/core/reader/types/element.rs
```diff
@@ -0,0 +1,165 @@
...
+            let use_of_el_ty_and_el_exprs = ty & 0b100 == 0b100;
+
+            let reftype_or_elemkind: Option<RefType> = match if ty & 0b011 != 0 {
+                if use_of_el_ty_and_el_exprs {
+                    Some(wasm.read_u8()?)
+                } else {
+                    let read = wasm.read_u8()?;
+                    match read {
+                        0x00 => None,
+                        _ => todo!("Only FuncRefs are allowed"),
```
nerodesu017, line 92: Might have to add full support for ExternRefs as well
Or should we wait until the Linker gets implemented?

src/core/reader/types/element.rs
```diff
@@ -0,0 +1,165 @@
...
+                init: items,
+                mode: elem_mode,
+            };
+
+            Ok(el)
+        })
+    }
+}
+
+/// Here we can't implement WasmReadable because we also want a mutable
```
nerodesu017, line 133: Oh, I won't forget to move this above!!!

src/execution/mod.rs
```diff
@@ -342,6 +344,260 @@ where
...
+
+                        assert!(table.len() >= (offset + el.len()));
+
+                        el.elem.iter().enumerate().for_each(|(i, rref)| {
+                            table.elem[i + offset] = *rref;
+                        });
+
+                        Some(el)
+                    }
+                    ElemMode::Declarative => {
```
nerodesu017, line 594: Declarative Segments are for streaming compilers only. As we don't do that, we can safely ignore them when executing.
WARNING: we can not fully get rid of them, we need to keep track of all elements, and if we don't, we're gonna break

src/execution/mod.rs
```diff
@@ -342,6 +344,260 @@ where
...
+                                            sub_programs.iter().map(|sub_program| {
+                                                let value = {
+                                                    let mut wasm = WasmReader::new(validation_info.wasm);
+                                                    wasm.move_start_to(*sub_program).unwrap_validated();
+                                                    let mut stack = Stack::new();
+                                                    // TODO: fully implement run_const
+                                                    run_const(wasm, &mut stack, (), &function_instances);
+                                                    let value = stack.peek_unknown_value();
+                                                    if value.is_none() {
+                                                        panic!("No value on the stack for element segment offset");
```
george-cosma, line 427: Ditto, error or unreachable_validated() or something like that

src/execution/mod.rs
```diff
@@ -342,6 +344,260 @@ where
...
+                                                    }
+                                                    _ => {
+                                                        unreachable!()
+                                                    },
+                                                };
+
+                                                Ref::Extern(ExternAddr::new(Some(offset as usize)))
+                                            }).collect::<Vec<Ref>>()
+                                        },
+                                        ElemItems::RefFuncs(_) => panic!("RefFuncs allowed only for Functions!")
```
george-cosma, line 456: Ditto panic removal

src/execution/mod.rs
```diff
@@ -342,6 +344,260 @@ where
...
+                                    }
+                                }
+                            },
+                        })
+                    },
+                    ElemMode::Active(active_elem) => {
+                        // copies itself right now, when instantiating
+                        let table_idx = active_elem.table as usize;
+                        assert!(tables.len() > table_idx);
```
george-cosma, line 465: I think assert_debug or an explicit check, to remove panics at runtime

src/execution/mod.rs
```diff
@@ -342,6 +344,260 @@ where
...
+
+                        let value = {
+                            let mut wasm = WasmReader::new(validation_info.wasm);
+                            wasm.move_start_to(active_elem.offset).unwrap_validated();
+                            let mut stack = Stack::new();
+                            // TODO: fully implement run_const
+                            run_const(wasm, &mut stack, (), &[]);
+                            let value = stack.peek_unknown_value();
+                            if value.is_none() {
+                                panic!("No value on the stack for element segment offset");
```
george-cosma, line 478: Ditto panics

src/execution/mod.rs
```diff
@@ -342,6 +344,260 @@ where
...
+                            value.unwrap()
+                        };
+
+                        // TODO: this shouldn't be a simple value, should it? I mean it can't be, but it can also be any type of ValType
+                        // TODO: also, do we need to forcefully make it i32?
+                        let offset: u32 = match value {
+                            Value::I32(val) => val,
+                            Value::I64(val) => {
+                                if val > u32::MAX as u64 {
+                                    panic!("i64 value for data segment offset is out of reach")
```
george-cosma, line 489: Ditto panic

src/execution/mod.rs
```diff
@@ -342,6 +344,260 @@ where
...
+                                                Ref::Extern(ExternAddr::new(Some(offset as usize)))
+                                            }).collect::<Vec<Ref>>()
+                                        },
+                                        ElemItems::RefFuncs(_) => panic!("RefFuncs allowed only for Functions!")
+                                    }
+                                }
+                            },
+                        })
+                    },
+                    ElemMode::Active(active_elem) => {
```
george-cosma, line 462: This feels quite similar to the implementation of ElemMode::Passive. Could we perhaps merge the two and the differences (if there aren't a lot) to be handled with a handful of if let / small nested match statements between the common code

src/execution/mod.rs
```diff
@@ -342,6 +344,260 @@ where
...
+                                                    let mut wasm = WasmReader::new(validation_info.wasm);
+                                                    wasm.move_start_to(*expr).unwrap_validated();
+                                                    let mut stack = Stack::new();
+                                                    // TODO: fully implement run_const
+                                                    run_const(wasm, &mut stack, (), &[]);
+                                                    let value = stack.peek_unknown_value();
+                                                    if value.is_none() {
+                                                        panic!("No value on the stack for element segment offset");
+                                                    }
+                                                    value.unwrap()
```
george-cosma, line 517: I'm not going to write out all "Ditto panic"s here, you get the idea. Also, I failed to notice until now the `.unwrap()`. May I suggest `.unwrap_validated()`

src/execution/value.rs
```diff
@@ -261,14 +261,141 @@ pub enum Value {
...
+    }
+}
+
+#[derive(Clone, Copy, PartialEq)]
+pub struct FuncAddr {
+    pub is_null: bool,
+    // it is the idx of the function in the current module
+    pub addr: usize,
+}
+
```
george-cosma, line 311: I know we discussed this in depth, but I think I might have changed my mind. Sorry :(

Okay, this isn't a blocking issue for me, but I would like to bring to the discussion making this a newtype
```rs
pub struct FuncAddr(Option<usize>)
```

This would add more "type safety", in the sense that a) a FuncAddress with a non-zero `addr` and true `is_null` shouldn't exist as a state in the code, and b) to avoid situations in which we forget to check `is_null`, the option variant is more safe

george-cosma (changes requested):
I haven't had the opportunity to review everything, but I've left some comments that you can ponder over tomorrow. Some are nitpicks, some are design choices.

I'd like to also highlight the fact that, at some point, we'll need to rework the constant expressions to be more expansive. It requires some linker (aka import/export) interfaces, though. But let's do this in anotther PR


I'll come back tomorrow/the day after tomorrow with more nitpicks and anything else relevant

src/execution/value.rs
george-cosma: Again, please feel free to push back on this. "It already works" is a valid argument. I'd just like to send it up for discussion

src/core/reader/types/mod.rs
```diff
@@ -85,6 +87,40 @@ pub enum RefType {
...
+        write!(
+            f,
+            "{}",
+            match self {
+                Self::ExternRef => "ExternRef",
+                Self::FuncRef => "FuncRef",
+            }
+        )
+    }
+}
```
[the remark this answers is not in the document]
nerodesu017: I couldn't find a way to derive Display without having to add the `derive_more` crate, a thing that we said we didn't really wanna do (add more crates) if possible.

src/execution/mod.rs
```diff
@@ -342,6 +344,260 @@ where
...
+        // https://webassembly.github.io/spec/core/syntax/modules.html#element-segments
+        let elements: Vec<ElemInst> = validation_info
+            .elements
+            .iter()
+            .enumerate()
+            .filter_map(|(i, el)| {
+                use crate::core::reader::types::element::*;
+                trace!("{:#?}", el);
+                match el.mode.clone() {
+                    ElemMode::Passive => {
```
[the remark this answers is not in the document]
nerodesu017: Good point

src/execution/value.rs
```diff
@@ -261,14 +261,141 @@ pub enum Value {
...
+        match self {
+            Self::Extern(extern_addr) => extern_addr.is_null,
+            Self::Func(func_addr) => func_addr.is_null,
+        }
+    }
+
+    pub fn is_specific_func(&self, func_id: u32) -> bool {
+        match self {
+            Self::Func(func_addr) => !func_addr.is_null && func_addr.addr == func_id as usize,
+            _ => unimplemented!(),
```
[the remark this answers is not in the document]
nerodesu017: Missed this one!

src/execution/mod.rs
```diff
@@ -342,6 +344,260 @@ where
...
+
+        let mut passive_elem_indexes: Vec<usize> = vec![];
+        // https://webassembly.github.io/spec/core/syntax/modules.html#element-segments
+        let elements: Vec<ElemInst> = validation_info
+            .elements
+            .iter()
+            .enumerate()
+            .filter_map(|(i, el)| {
+                use crate::core::reader::types::element::*;
+                trace!("{:#?}", el);
```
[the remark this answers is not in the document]
nerodesu017: Good one

src/execution/mod.rs
nerodesu017: Modifying to an unreachable

src/execution/mod.rs
nerodesu017: This is a tad more complicated. At the moment, when instantiating the store we simply return a Store when, I think, we should return a Result<Store>. The Error should probably be apart of a StoreInstantiationError enum, and thus even the parent (new_with_hooks) return not just a Result<Self, RuntimeError> but a Result<Self, RuntimeError | StoreInstantiationError>

We shall discuss this on Wednesday

src/execution/mod.rs
```diff
@@ -431,6 +591,40 @@ where
...
+    span: &Span,
+    imported_globals: (),
+    imported_funcs: &[FuncInst],
+) -> Value {
+    let mut wasm = WasmReader::new(wasm);
+    wasm.move_start_to(*span).unwrap_validated();
+    let mut stack = Stack::new();
+    run_const(wasm, &mut stack, imported_globals, imported_funcs);
+    let value = stack.peek_unknown_value();
+    if value.is_none() {
```
nerodesu017, line 626: Might want to change this function to return an Error instead of just panic (related to George's previous comments). Maybe a StoreInstantiationError?

src/core/reader/types/mod.rs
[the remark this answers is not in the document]
nerodesu017: Do NOT implement display !!!1! (don't forget the 1)
Remove it.

src/execution/mod.rs
```diff
@@ -364,7 +524,7 @@ where
                         let mut wasm = WasmReader::new(validation_info.wasm);
                         wasm.move_start_to(active_data.offset).unwrap_validated();
                         let mut stack = Stack::new();
-                        run_const(wasm, &mut stack, ());
+                        run_const(wasm, &mut stack, (), &[]);
                         let value = stack.peek_unknown_value();
                         if value.is_none() {
```
nerodesu017, line 529: I think we have a value when validating. Can we unwrap_validated here?

src/execution/mod.rs
```diff
@@ -342,6 +344,260 @@ where
...
+        // https://webassembly.github.io/spec/core/exec/modules.html#tables
+        let mut tables: Vec<TableInst> = validation_info
+            .tables
+            .iter()
+            .map(|ty| TableInst::new(*ty))
+            .collect();
+
+        let mut passive_elem_indexes: Vec<usize> = vec![];
+        // https://webassembly.github.io/spec/core/syntax/modules.html#element-segments
+        let elements: Vec<ElemInst> = validation_info
```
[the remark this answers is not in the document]
nerodesu017: We have to split this down further, maybe yet another (or more) function(s)?

src/execution/mod.rs
nerodesu017: Might be able to unwrap_validated, check for it

src/execution/value.rs
nerodesu017: Let's go for the Option choice. Might be a lot of rework, but might also be much easier to understand.

nerodesu017: Use just a USIZE
If u32::MAX => NULL
Do it like wasmtime B)

nerodesu017: Let's check the spec on this

nerodesu017: with the Option!!!!! 🤔

src/execution/value.rs
```diff
@@ -261,14 +261,161 @@ pub enum Value {
...
+    }
+}
+
+impl Default for FuncAddr {
+    fn default() -> Self {
+        Self::null()
+    }
+}
+
+impl Display for FuncAddr {
```
nerodesu017, line 355: you might want to REMOVE this

src/execution/value.rs
```diff
@@ -261,14 +261,161 @@ pub enum Value {
...
+    }
+}
+
+impl Default for ExternAddr {
+    fn default() -> Self {
+        Self::null()
+    }
+}
+
+impl Display for ExternAddr {
```
nerodesu017, line 405: this as well

src/execution/const_interpreter_loop.rs
```diff
@@ -95,6 +101,23 @@ pub(crate) fn run_const(
...
                 stack.push_value(res.into());
             }
+            REF_NULL => {
+                let reftype = RefType::read_unvalidated(&mut wasm);
+
+                stack.push_value(Value::Ref(reftype.to_null_ref()));
+                trace!("Instruction: ref.null '{}' -> [{}]", reftype, reftype);
+            }
+            REF_FUNC => {
+                // not unwrap validated because, guess what? not validated
```
george-cosma, line 111: Why is this not validated?

src/execution/mod.rs
```diff
@@ -431,6 +591,40 @@ where
...
+        // INFO: from wasmtime - implement only global
+        _ => unreachable!(),
+    }
+}
+
+fn read_value_from_wasm(
+    wasm: &[u8],
+    span: &Span,
+    imported_globals: (),
+    imported_funcs: &[FuncInst],
```
george-cosma, line 619: Why "imported_funcs"? Are elements intializable only with imported funcs?

src/execution/mod.rs
```diff
@@ -431,6 +591,40 @@ where
...
+            elements,
+            passive_elem_indexes,
         }
     }
 }
+
+fn get_address_offset(value: Value) -> u32 {
+    match value {
+        Value::I32(val) => val,
+        Value::I64(val) => {
```
george-cosma, line 604: Can you provide a source that attests that only I32s and I64 are generated by the initialiation expression of elements?

george-cosma: Also, some source for the weird I64 behaviour would be nice

george-cosma:
This is on your repo so I can't add a commit to it. As such, we'll do it 90's style and give you the patch file to apply, if you find it correct
Note: file renamed from `proposal.patch` to `proposal.txt` to be able to upload to github... because it filters based on file ending and not contents. Security Theater!

[proposal.txt](https://github.com/user-attachments/files/17939722/proposal.txt)

src/execution/const_interpreter_loop.rs
nerodesu017: You are right, it should be validated because we do validation first using `read_constant_instructions`

src/execution/mod.rs
```diff
@@ -342,6 +345,90 @@ where
...
+                        .iter()
+                        .map(|expr| {
+                            get_address_offset(
+                                run_const_span(validation_info.wasm, expr, ()).unwrap_validated(),
+                            )
+                        })
+                        .collect(),
+                    ElemItems::RefFuncs(indicies) => {
+                        // For external references, this branch will never be taken
+                        // TODO: REASON @nerodesu017
```
nerodesu017, line 375: for me TODO: add comments

src/execution/mod.rs
nerodesu017: I64 was wrong. Not even wasmtime implements it. They do that, though, for data section (as they implement the [Memory64 Proposal](https://github.com/WebAssembly/memory64)).

george-cosma (commented):
One more nitpick. The code under src/core/reader/types/element.rs seems to do more than just parsing.
I now understand why you were asking about WasmReader in the lass call.

All code under core/reader(/types) is for parsing, no validation or execution
All code under execution/ is for .... execution
All code under validation/ is for validation

This is my (implicit) understanding of the code structure. From what I can the element.rs file contains more than that - it also does some validation. I would suggest we split it up - we first parse what there is to parse (no validation), and under validation we verify what we parsed. Similar to globals.

george-cosma:
[patch3.txt](https://github.com/user-attachments/files/17994589/patch3.txt)

src/execution/value.rs
```diff
@@ -261,14 +261,109 @@ pub enum Value {
...
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        match self {
+            Ref::Func(func_addr) => write!(f, "FuncRef({:?})", func_addr),
+            Ref::Extern(extern_addr) => write!(f, "ExternRef({:?})", extern_addr),
+        }
+    }
+}
+
+#[derive(Clone, Copy, PartialEq)]
+pub struct FuncAddr {
```
nerodesu017, line 306: Doc String here to reference the wasm tables?
Maybe new types? like `FuncIdx = usize`

src/execution/value.rs
```diff
@@ -261,14 +261,109 @@ pub enum Value {
...
+        Self::null()
+    }
+}
+
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub struct ExternAddr {
+    pub addr: Option<usize>,
+}
+
+impl ExternAddr {
```
nerodesu017, line 345: Doc String or new type here, as well

src/core/reader/types/element.rs
```diff
@@ -0,0 +1,164 @@
...
+    core::reader::span::Span, read_constant_expression::read_constant_instructions, Error, Result,
+};
+
+use super::RefType;
+
+#[derive(Clone)]
+pub struct ElemType {
+    pub init: ElemItems,
+    pub mode: ElemMode,
+}
```
cemonem, line 15: I am strongly in favor of this to be in this form for better correspondence with the spec and understandability:
```
pub struct ElemSegment {
  pub ty: RefType,
  pub init: Vec<Span>,
  pub mode: ElemMode
}
```
This can be created during element section read of validation::validate() and also validated.
and then the bunch of ElemSegments could be instantiated to ElemInst's within Store::init_store() (with the const spans evaluated), almost exactly like Globals to GlobalInsts. I think treating them like a mixture of DataSegments and Globals within our implementation feels like a such a perfect fit, I would feel bad glossing over it.

george-cosma: I'm not 100% convinced that this issue is critical enough for the effort needed to refactor this. Also, having to convert u32 (in the case of reffuncs) into initialization expressions seems inefficient.
- For `ty = RefFuncs`, we  can have the indicies directly in the binary `u32` or as expressions, which isn't really represented your proposal.
- For `ty = ExternFuncs` we only have expressions (afaik).

src/core/reader/types/element.rs
```diff
@@ -0,0 +1,164 @@
...
+                } else {
+                    ElemMode::Passive
+                }
+            } else {
+                let table_idx = if ty & 0b010 == 0b010 {
+                    wasm.read_var_u32()?
+                } else {
+                    0
+                };
+                if tables_length <= table_idx as usize {
```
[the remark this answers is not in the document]
nerodesu017: issue #102

src/core/reader/types/element.rs
cemonem: fair enough :)

nerodesu017:
I have disabled testsuite automatic testing because we... panic

cemonem (approved):

PR #101, opened 2024-12-04
structured ctrl

cemonem:
### Pull Request Overview

- implementations for block...end, br, br_if, br_table, return, loop...end, if..else...end instructions.
- ctrl stack and type stack are separated during validation to avoid linear look up for ctrl labels.
- sidetable generation per function.
- integration of sidetable and sidetable pointer in the interpreter loop for all ctrl structures.
- execution tests for structured_ctrl.
- assertion of types in the type stack during validation might mutate the stack. Any error during validation may leave the stack in a modified, invalid state.
- validation tests for cases where assertion of types that do mutate the stack, and this change is required.
- preparation for the type polymorphic select instruction (Polymorphic type NumOrVecType for select instruction on the type stack)

### TODO
- Per module sidetables instead of per function sidetables.
- Backpatching sidetable entries without extra memory / without vectors in the CtrlStackEntries (using the linked list trick in wizard)

### Formatting

- [x] Ran `cargo fmt`
- [x] Ran `cargo check`
- [x] Ran `cargo build`
- [x] Ran `cargo doc`
- [x] Ran `nix fmt`


src/core/reader/types/values.rs
```diff
@@ -67,6 +67,27 @@ impl WasmReader<'_> {
...
+                break;
+            }
+        }
+
+        if shift < 33 && (byte & 0x40 != 0) {
+            result |= !0 << shift;
+        }
+
+        Ok(result)
+    }
```
george-cosma, line 89: Are we confident this isn't the same as `read_var_i32` in practice? Since the first bit will always be 0 to be positive, do we need to store this in a i64?

src/core/reader/mod.rs
```diff
@@ -211,6 +211,11 @@ pub mod span {
         pub const fn len(&self) -> usize {
             self.len
         }
+
+        // TODO is this ok?
+        pub const fn from(&self) -> usize {
+            self.from
+        }
```
george-cosma, line 218: 👍

src/core/reader/types/mod.rs
```diff
@@ -201,6 +201,86 @@ impl WasmReadable for FuncType {
...
+    Returns(ValType),
+    Type(u32),
+}
+
+impl WasmReadable for BlockType {
+    fn read(wasm: &mut WasmReader) -> Result<Self> {
+        // FIXME: Use transactions for ValType::read
+        if wasm.peek_u8()? as i8 == 0x40 {
+            // Empty block type
+            let _ = wasm.read_u8().expect("read to succeed, as we just peeked");
```
george-cosma, line 217: `.unwrap_validated` imho, you can provide the reason as a comment, though it is pretty self-explanatory

src/core/reader/types/mod.rs
```diff
@@ -201,6 +201,86 @@ impl WasmReadable for FuncType {
...
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum BlockType {
+    Empty,
+    Returns(ValType),
+    Type(u32),
+}
+
+impl WasmReadable for BlockType {
+    fn read(wasm: &mut WasmReader) -> Result<Self> {
+        // FIXME: Use transactions for ValType::read
```
george-cosma, line 214: It's okay with or without. I don't care.

src/validation/mod.rs
```diff
@@ -30,7 +31,7 @@ pub struct ValidationInfo<'bytecode> {
     pub(crate) globals: Vec<Global>,
     #[allow(dead_code)]
     pub(crate) exports: Vec<Export>,
-    pub(crate) func_blocks: Vec<Span>,
+    pub(crate) func_blocks_sidetables: Vec<(Span, Sidetable)>,
```
george-cosma, line 34: I think we can keep the name `func_blocks` and use a docstring to explain that we also attach the sidetable

src/execution/interpreter_loop.rs
```diff
@@ -64,7 +70,20 @@ pub(super) fn run<H: HookSet>(
...
+
+                // TODO there is definitely a better to write this
+                let current_func_span = store
+                    .funcs
+                    .get(stack.current_stackframe().func_idx)
+                    .unwrap_validated()
+                    .code_expr;
+                if wasm.pc != current_func_span.from() + current_func_span.len() {
+                    continue;
+                }
```
george-cosma, line 84: Is there a reason why we continue and we don't break?

```suggestion
                let current_func_span = store
                    .funcs
                    .get(stack.current_stackframe().func_idx)
                    .unwrap_validated()
                    .code_expr;

                if wasm.pc != current_func_span.from() + current_func_span.len() {
                    continue;
                }
```

src/execution/interpreter_loop.rs
```diff
@@ -75,30 +94,36 @@ pub(super) fn run<H: HookSet>(
...
-                }
+                let c: i32 = stack.pop_value(ValType::NumType(NumType::I32)).into();
 
-                if stack.callframe_count() == 1 {
-                    break;
+                if c != 0 {
+                    do_sidetable_control_transfer(&mut wasm, stack, &mut stp, current_sidetable);
+                } else {
+                    stp += 1;
                 }
```
george-cosma, line 114:
```suggestion
                let test_val: i32 = stack.pop_value(ValType::NumType(NumType::I32)).into();

                if test_val != 0 {
                    do_sidetable_control_transfer(&mut wasm, stack, &mut stp, current_sidetable);
                } else {
                    stp += 1;
                }
```

src/execution/interpreter_loop.rs
```diff
@@ -75,30 +94,36 @@ pub(super) fn run<H: HookSet>(
...
-                trace!("end of function reached, returning to previous stack frame");
-                wasm.pc = stack.pop_stackframe();
+            }
+            BR => {
+                //skip n of BR n
+                wasm.read_var_u32().unwrap_validated();
+                do_sidetable_control_transfer(&mut wasm, stack, &mut stp, current_sidetable);
+            }
+            BLOCK | LOOP => {
+                BlockType::read_unvalidated(&mut wasm);
```
george-cosma, line 122: Arw we doing anything with this return value?

george-cosma (changes requested):
My review is not yet done, I still have to check the validation stack changes

src/core/reader/types/values.rs
cemonem: yeah they turned out to be different. We can't use i32 because the value inside might be larger than 2^31-1. We store it in an i64 in case the encoding turns out to be larger than u32 or negative, which we throw an error.

src/execution/interpreter_loop.rs
cemonem: if this is not the very last END instr, END does not do anything, so we read the next instr (wasm.pc < last_end_idx). The dodgy case here is when the pc is erroneously after the very last END (wasm.pc > last_end_idx). I let it continue here, but we could also give an error or trap saying "instruction pointer is out of bounds"

src/execution/interpreter_loop.rs
cemonem: No, I simply parsed it to skip it since blocktype parsing is a bit convoluted, but I can write a skip method as well.

src/validation/mod.rs
cemonem: Per function sidetables are going to be replaced with per module sidetables anyway.

src/core/reader/types/mod.rs
```diff
@@ -201,6 +201,85 @@ impl WasmReadable for FuncType {
...
+            Ok(BlockType::Empty)
+        } else if let Ok(val_ty) = ValType::read(wasm) {
+            // No parameters and given valtype as the result
+            Ok(BlockType::Returns(val_ty))
+        } else {
+            // An index to a function type
+            wasm.read_var_i33()
+                .and_then(|idx| idx.try_into().map_err(|_| Error::InvalidFuncTypeIdx))
+                .map(BlockType::Type)
+        }
```
florianhartung, line 226: I think there might be an error here, coming from a general issue with the current WasmReader-pattern.
When `ValType::read` is called, it might consume some bytes and only then realize the input is invalid. Thus it returns `Err(_)`, however the previously consumed bytes are never returned to the `WasmReader`.
In this case this code would then go into the `else` block and try to read an `i33` from the now different starting index.

In my implemention of this method I added the comment: `// FIXME: Use transactions for ValType::read`, which is missing here. I suggest something along the lines of:
```suggestion
        if wasm.peek_u8()? as i8 == 0x40 {
            // Empty block type
            let _ = wasm.read_u8().unwrap_validated();
            Ok(BlockType::Empty)
        } else if let Ok(val_ty) = wasm.handle_transaction(|wasm| ValType::read(wasm)) {
            // No parameters and given valtype as the result
            Ok(BlockType::Returns(val_ty))
        } else {
            // An index to a function type
            wasm.read_var_i33()
                .and_then(|idx| idx.try_into().map_err(|_| Error::InvalidFuncTypeIdx))
                .map(BlockType::Type)
        }
```

florianhartung (changes requested):

src/core/reader/types/mod.rs
florianhartung: Please do this for both `BlockType::read` and `BlockTyp::read_unvalidated`

florianhartung (changes requested):

src/core/reader/types/mod.rs
florianhartung: Please do the previous change for both `BlockType::read` and `BlockTyp::read_unvalidated`

src/core/reader/types/values.rs
cemonem: Another reason they are using i encoding instead of u encoding is that unsigned single byte numbers might clash with the binary encoding of NumType etc. If we force that numbers larger than 2^7-1 are negative if its a single byte (first bit will be always 0 since that's how leb128 works) even if their unsigned repr can fit in a single byte, we enforce the blocktype to be at least two bytes.

wucke13 (approved):

wucke13:
@cemonem Please address my two final remarks and the one from george, and then lets merge!

florianhartung:
Congrats on getting the sidetable and this MR finished @cemonem 🥳👍

Now that our interpreter has proper control flow, one question remains: Will it run Doom? /s

george-cosma (approved):

src/execution/value_stack.rs
```diff
@@ -210,10 +213,16 @@ impl Stack {
     }
 
     /// Clear all of the values pushed to the value stack by the current stack frame
+    #[allow(unused)] // TODO remove this once sidetable implementation lands
```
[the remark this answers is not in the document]
george-cosma: @cemonem are you sure you force-pushed everything you had locally? This doesn't seem to be resolved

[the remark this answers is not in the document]
cemonem: yeah you're right sorry, this last one wasn't even staged, my bad.

wucke13 (approved):

wucke13 (approved):

PR #103, opened 2024-12-09
Improve whitepaper code and small fixes

florianhartung:
### Pull Request Overview

This pull request changes the structure of out whitepaper Typst project.
I have used Typst myself and this pull request contains some opinionated improvements and fixes for our project.

### Formatting

- [x] Ran `cargo fmt`
- [x] Ran `cargo check`
- [x] Ran `cargo build`
- [x] Ran `cargo doc`
- [x] Ran `nix fmt`

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -196,11 +196,11 @@
     "typst-packages": {
       "flake": false,
       "locked": {
-        "lastModified": 1731485222,
-        "narHash": "sha256-uhXDznbGLS4s6d2S8OHUCzpgmJQ1H2HCJug7kRIY0Jw=",
+        "lastModified": 1733827512,
+        "narHash": "sha256-++W3zUwRreWSZmPP/sUTpQZjOrj5chnQZhlmePm521M=",
         "owner": "typst",
         "repo": "packages",
-        "rev": "4979ee3923fb0143e5d076854cb0d94bbf367a2c",
+        "rev": "dfd4a3ee4a819b11640361610bc0a3eb10cd64e7",
         "type": "github"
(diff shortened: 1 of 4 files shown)


florianhartung:
Why is the automatic deployment not working?
For me https://packages.typst.org/preview/acrostiche-0.4.1.tar.gz works just fine.

george-cosma:
@florianhartung I think I know what the issue is. Your PR would work with acrostiche 0.4.0, but not with 0.4.1 . Nix is a "reproducible" built system so it builds everything from scratch (+ cache), but web requests are not reproducible, so internet is disabled in the build process.

My hypothesis is that the package 0.4.0 is indexed by <<whatever is responsible to manage typst packages in nix>> and build in-place, but 0.4.1 is not. As a fail-safe it tries to download it from the internet but since Nix is Nix, it blocks web requests and thus it fails.

This is most likely an issue with the nix script, a nix dependency versioning issue, or the nix packages itself that handles typst packages.  @wucke13 would have a better intuition in this sense.

The short answer: downgrade to acrostiche 0.4.0 (like it was originally). I will try to look into it, but not guarantees to make it work.

florianhartung:
@george-cosma Yes you were correct, the flake input `typst-packages` was still locked to an old version where `acrostiche 0.4.1` didn't exist yet. I fixed it by running:

```console
nix flake lock --update-input typst-packages
```

wucke13 (approved):

cemonem (approved):

george-cosma (approved):

Issue #28, opened 2024-07-10
Info: Some architectures alter floats


nerodesu017:
As said in <https://github.com/bytecodealliance/wasmtime/blob/7caca1624a3c68dd4135b3941e7c87d94eb4c7db/cranelift/codegen/src/ir/immediates.rs#L471>

> We specifically avoid using a f16 here since some architectures may silently alter floats

So we should handle floats by having u32 and u64 wrapped in structs

florianhartung:
Maybe this could be a viable solution for now:
```rs
pub struct Ieee32(pub f32);
// impls for floating point operations which just make calls to the underlying f32
```

This way we can save some time now while still being able to implement the float operations ourselves later on 🤔

PR #112, opened 2024-12-16
Official Testsuite Runner

nerodesu017:
### Pull Request Overview

This is an on-going PR focused on making the spec test runner work again + add more to it. 
Additions:
  - filtering (either Include or Exclude) based on files and folders (so we can discard the `proposals` folder, for example, or files related to `SIMD`, which we do not implement (yet), and that give us a rough time)
  - `assert_trap` being implemented
  - a few `debug` (with `debug_assertions`) related code for stringifying errors and other features for QoL when debugging
  - changes to how we cast certain values (especially F64 and F32) so that we get the desired results (mostly bit reinterpretation, not normal casting, this is in part because the crate that we use for parsing WAST files parses these values into u32 and u64, respectively, and we have to reinterpret the bits - see #28 for more info)
  - a fix for a Span that we return when reading a constant expression (for example, for a global) because we were returning 1 more byte, this wouldn't be a problem, as we stop on the `END` instruction, but it is a problem when the `END` is EOF and we return +1 byte and we throw an error when doing a later check - see [src/validation/read_constant_expression.rs](https://github.com/DLR-FT/wasm-interpreter/pull/112/files#diff-2d81037229a0aec48ea1dc6159d1bef8f41fcd01795b38efb9df7520305a0115)

See commit message for more!

### Testing Strategy

This pull request was tested by... N/A

### TODO or Help Wanted

This pull request still needs... N/A

### Formatting

- [x] Ran `cargo fmt`
- [x] Ran `cargo check`
- [x] Ran `cargo build`
- [x] Ran `cargo doc`
- [ ] Ran `nix fmt`

### Github Issue


nerodesu017:
Can't use harness with [[test]] in Cargo.toml because nix action will whine about output.
Also, can't really upload them anywhere because github actions do NOT want to work with me here and it's kind of painful

cemonem (approved):

george-cosma (approved):

cemonem (approved):

george-cosma (approved):

george-cosma:
@nerodesu017 I'll let you merge (in the morning)

PR #113, opened 2025-01-19
Dev/linker rebase

george-cosma:
### Pull Request Overview

### Testing Strategy

### TODO or Help Wanted

### Formatting

- [ ] Ran `cargo fmt`
- [ ] Ran `cargo check`
- [ ] Ran `cargo build`
- [ ] Ran `cargo doc`
- [ ] Ran `nix fmt`

### Github Issue

### DIFF
--- a/src/validation/code.rs
+++ b/src/validation/code.rs
@@ -23,6 +23,7 @@ pub fn validate_code_section(
     section_header: SectionHeader,
     fn_types: &[FuncType],
     type_idx_of_fn: &[usize],
+    num_imported_funcs: usize,
     globals: &[Global],
     memories: &[MemType],
     data_count: &Option<u32>,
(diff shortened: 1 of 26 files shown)


src/validation/code.rs
```diff
@@ -335,7 +339,7 @@ fn read_instructions(
                         // the last end instruction will handle the return to callee during execution
                         stps_to_backpatch.iter().for_each(|i| {
                             sidetable[*i].delta_pc =
-                                (wasm.pc as isize) - sidetable[*i].delta_pc - 1; // minus 1 is important!
+                                (wasm.pc as isize) - sidetable[*i].delta_pc - 1; // minus 1 is important! TODO: Why?
```
cemonem, line 342: Because we want to execute the very last end instruction of the function block no matter what, as it triggers the non-jump return procedures (pop the call stack etc.). The other end instructions are essentially nops during execution, so we jump past them and simply execute the next instruction, not the associated end instruction itself.

george-cosma:
Force pushed into #87

src/validation/code.rs
george-cosma: Leave this comment on the main PR. I forced pushed this into it

PR #117, opened 2025-01-20
select instruction validation/execution

cemonem:
### Pull Request Overview

This pull request implements the polymorphic and non-polymorphic variants of the SELECT instruction, and adds some tests.

### Formatting

- [x] Ran `cargo fmt`
- [x] Ran `cargo check`
- [x] Ran `cargo build`
- [x] Ran `cargo doc`
- [x] Ran `nix fmt`


src/validation/validation_stack.rs
```diff
@@ -320,6 +344,34 @@ pub enum ValidationStackEntry {
...
+
+    fn unify_to_num_or_vec_type(&self) -> Result<Self> {
+        match self {
+            ValidationStackEntry::Val(ValType::NumType(_)) => Ok(self.clone()),
+            ValidationStackEntry::Val(ValType::VecType) => Ok(self.clone()),
+            ValidationStackEntry::NumOrVecType => Ok(self.clone()),
+            ValidationStackEntry::UnspecifiedValTypes => Ok(ValidationStackEntry::NumOrVecType),
+            _ => Err(Error::TypeUnificationMismatch),
+        }
+    }
```
george-cosma, line 372: This logic makes more sense to me directly put into unify

src/validation/validation_stack.rs
```diff
@@ -320,6 +344,34 @@ pub enum ValidationStackEntry {
...
+                Self::Val(o) => (o == s)
+                    .then_some(self.clone())
+                    .ok_or(Error::TypeUnificationMismatch),
+                Self::NumOrVecType => self.unify_to_num_or_vec_type(),
+                Self::UnspecifiedValTypes => Ok(self.clone()),
+            },
+            ValidationStackEntry::NumOrVecType => other.unify_to_num_or_vec_type(),
+            ValidationStackEntry::UnspecifiedValTypes => Ok(other.clone()),
+        }
+    }
```
george-cosma, line 362: Not really necessary to have a seperate function that is only called from one place.
If we don't forsee this being used elsewhere, a simple comment+code inside the calling function would suffice

// Try to unify the values
<code for the unify here>

src/validation/validation_stack.rs
```diff
@@ -320,6 +344,34 @@ pub enum ValidationStackEntry {
...
 }
+
+impl ValidationStackEntry {
+    // TODO better way to write this?
+    fn unify(&self, other: &ValidationStackEntry) -> Result<Self> {
+        match self {
+            ValidationStackEntry::Val(s) => match other {
+                Self::Val(o) => (o == s)
+                    .then_some(self.clone())
+                    .ok_or(Error::TypeUnificationMismatch),
```
george-cosma, line 355: The `(o == s).then_some` syntax is a bit ... esoteric for me. Not a big issue but I would prefer a simple if else :)

src/validation/validation_stack.rs
```diff
@@ -302,6 +302,30 @@ impl ValidationStack {
...
+        //SELECT instruction has the type signature
+        //[t t i32] -> [t] where t is a Num or Vec Type
+
+        // TODO write this more efficiently
+        self.assert_pop_val_type(ValType::NumType(crate::NumType::I32))?;
+
+        let unified = self
+            .pop_valtype()?
+            .unify(&self.pop_valtype()?)
+            .map_err(|_| Error::InvalidValidationStackValType(None))?;
```
george-cosma, line 316: Why `None`? Report one or the other as invalid

src/validation/validation_stack.rs
```diff
@@ -302,6 +302,30 @@ impl ValidationStack {
             last_ctrl_stack_entry.block_ty,
         ))
     }
+
+    pub fn validate_polymorphic_select(&mut self) -> Result<()> {
+        //SELECT instruction has the type signature
+        //[t t i32] -> [t] where t is a Num or Vec Type
+
+        // TODO write this more efficiently
```
george-cosma, line 310: LGTM

src/validation/validation_stack.rs
```diff
@@ -320,6 +344,34 @@ pub enum ValidationStackEntry {
     // TODO change this name to BottomType
     UnspecifiedValTypes,
 }
+
+impl ValidationStackEntry {
+    // TODO better way to write this?
```
george-cosma, line 349: I would suggest we return Result<&Self> , those clones are eating away at the performance. I'm not going to go into the weeds, but i did a quick performance test and we are not sitting pretty...

src/validation/validation_stack.rs
```diff
@@ -302,6 +302,30 @@ impl ValidationStack {
...
+        self.assert_pop_val_type(ValType::NumType(crate::NumType::I32))?;
+
+        let unified = self
+            .pop_valtype()?
+            .unify(&self.pop_valtype()?)
+            .map_err(|_| Error::InvalidValidationStackValType(None))?;
+
+        match unified {
+            ValidationStackEntry::UnspecifiedValTypes => {
+                //if unified is a bottom type only way to satisfy validation of SELECT is to unify it to NumOrVec
```
george-cosma, line 320: Also, I would prefer we come to a conclusion upon the naming of the "can be any type" type. Having it UnspecifiedValueType in the code, but the bottom type in other parts is a tad confusing.

src/validation/validation_stack.rs
cemonem: How about simply letting ValidationStackEntries have Copy trait?

src/validation/validation_stack.rs
cemonem: Pseudocode calls it "Unknown/UnknownType". Wizard calls it "Bottom/BottomType". I'm fine with any of these choices.

src/validation/validation_stack.rs
cemonem: unification code is essentially an nxn matrix, which is quite ugly to write within matches within matches within ifs :)

src/validation/validation_stack.rs
cemonem: It might be used within assert_val_types_with_custom_stacks, I thought the same myself, its a bit hard for me to infer before the refactor.

george-cosma (approved):

src/validation/validation_stack.rs
cemonem: Let's leave Copy trait change for another pr.

src/validation/validation_stack.rs
cemonem: Let's leave the name change for the refactor pr.

src/validation/validation_stack.rs
cemonem: yeah writing it as if has the same readability :) done

george-cosma (approved):

wucke13 (approved):

PR #118, opened 2025-01-20
fix: per module sidetables

### LINKED ISSUE #115: sidetables should be generated per module, not per function

cemonem:
### Pull Request Overview

per module sidetables instead of per function, sidetables are stored within the Store object.

### Formatting

- [x] Ran `cargo fmt`
- [x] Ran `cargo check`
- [x] Ran `cargo build`
- [x] Ran `cargo doc`
- [x] Ran `nix fmt`

### Github Issue

This pull request closes #115

### DIFF
--- a/src/execution/interpreter_loop.rs
+++ b/src/execution/interpreter_loop.rs
@@ -54,10 +54,8 @@ pub(super) fn run<H: HookSet>(
     // Start reading the function's instructions
     let mut wasm = &mut modules[*current_module_idx].wasm_reader;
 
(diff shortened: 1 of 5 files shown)


cemonem:
This will probably have to be rewritten when the linker is merged in. what "current" sidetable is will still need to be held in the call stack, as we need to remember which module the caller was in, and if we call or return from a different module, we'll need to update the current sidetable.

george-cosma:
Is there a specific reason why we want to do this? Genuinely curious

cemonem:
has gone stalem, reimplemented within #131 .

PR #125, opened 2025-02-04
Nero/dev/runner upgraded

nerodesu017:
### Pull Request Overview

### Testing Strategy

### TODO or Help Wanted

### Formatting

- [ ] Ran `cargo fmt`
- [ ] Ran `cargo check`
- [ ] Ran `cargo build`
- [ ] Ran `cargo doc`
- [ ] Ran `nix fmt`

### Github Issue


.github/workflows/rust-test-runner.yml
george-cosma, line 1: Pushing to gh-pages is dangerous. It will 100% get forced-pushed over by the preview commits. I suggest we have the report generate in the /tests or root directory, and we push it up along other changes. Two actions would be needed: one that checks that the PR's current json is up to date (so we don't have bad data) and one that downloads the report from main. There's something to be said about comparing to the destination branch rather than main, but that's more of a hassle than needed.

tests/specification/mod.rs
george-cosma, line 1: If it doesn't cause the entire program to recompile, I would put the writing-thing under a feature flag. A more cool approach will be to take a hold of the testing harness and add a CLI flag, though that might be more complicated.

Cargo.toml
george-cosma, line 1: Just as a side-note I hope we won't have the do tool certification for serde, cause that would be pain. I'm banking on the fact that we use this only as a CI warning and is not crucial in the running of the testsuite

.github/workflows/rust-test-runner.yml
george-cosma: Making a separate branch for this would also be okay-ish. I'm not 100% sure that doing direct git commands inside of the CI action is best practice, but I don't know either way. Just a gut feeling that might be unsubstantiated.

.github/workflows/rust-test-runner.yml
```diff
@@ -0,0 +1,56 @@
...
+          path: gh-pages
+
+      - name: Fetch previous testsuite_results.json
+        run: |
+          cp gh-pages/main/testsuite_results.json testsuite_results_original.json
+
+      - name: Run tests
+        run: cargo test --package wasm-interpreter --test wasm_spec_testsuite --verbose -- --nocapture
+
+      - name: Save testsuite_results.json to gh-pages
```
george-cosma, line 40: I think this will overwrite the results file every time there is a new commit on the PR, and if there are 2 seperate PRs going on, they will intervene with one another. I think the best bet is to have another action that does the saving only when main has a new push

florianhartung:
closed in favor of https://github.com/DLR-FT/wasm-interpreter/pull/126

PR #126, opened 2025-02-08
feat: Testsuite Preview github action

george-cosma:
### Pull Request Overview

This pull request adds a sticky comment that shows differences between testsuite runs. For example, see https://github.com/george-cosma/wasm-interpreter/pull/11

Based on @nerodesu017 implementation #125 

### Testing Strategy

This pull request was tested here: https://github.com/george-cosma/wasm-interpreter/pull/11

### TODO or Help Wanted

Just as a note, when I tried modifying I32 sub the interpreter ran into an infinite loop. This should be investigated.

Also, this does add a python script. I will want to make another script that tests the speed of our interpreter against a reference program. This script will need to compile another rust program, and storing it in main seems.... weird. Suggestions/Is this a non-issue?

### Formatting

- [x] Ran `cargo fmt`
- [x] Ran `cargo check`
- [x] Ran `cargo build`
- [ ] Ran `cargo doc`
- [ ] Ran `nix fmt`

### DIFF
--- a/.github/workflows/testsuite_preview.yaml
+++ b/.github/workflows/testsuite_preview.yaml
@@ -1,0 +1,48 @@
+name: Testsuite Preview
+
+on:
+  pull_request:
+
+permissions:
+  pull-requests: write
+
+jobs:
+  ts_preview:
+    strategy:
+      matrix:
+        os: [ubuntu-latest]
+    runs-on: ${{ matrix.os }}
+    env:
+      RUSTFLAGS: "-D warnings"
+      RUSTDOCFLAGS: "-D warnings"
+
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          ref: main
+          submodules: true
+
+      - name: Generate old report
+        run: |
+          TESTSUITE_SAVE=1 cargo test -- spec_tests --show-output
+          cp testsuite_results.json old.json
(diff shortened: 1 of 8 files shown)


george-cosma:
Note, current check fails due to main branch not producing a `testsuite_result.json` file

cemonem:
Looks great, I think it's looks fair enough as far as seperation of concerns are concerned at the very beginning, we can change things up later. Once nix deps are added we can merge I think.

cemonem:
> I'm not sure what you mean by "seperation of concerns" or what "nix dependencies" are referencing

By seperation of concerns i mean what the python script does, where in the python script, or rust tests etc. By nix deps I meant the packages that have to be mentioned in the flake as a test dependency / devshell for the python script

wucke13:
I understand that the python script has no external dependencies, simply putting pkgs.python3 into the devshell could work already?

cemonem:
The contents of the devshell command I added looks bad but fair enough i guess, if there are better ways to write it, be my guest :sweat_smile:

cemonem (approved):

george-cosma:
Please see tracking PR #150

florianhartung:
this can be closed when https://github.com/DLR-FT/wasm-interpreter/pull/163 gets merged

PR #127, opened 2025-02-14
fix: wasm official testsuite bug hunt

cemonem:
### Pull Request Overview

so far return type check was fixed.

### Formatting

- [x] Ran `cargo fmt`
- [x] Ran `cargo check`
- [x] Ran `cargo build`
- [x] Ran `cargo doc`
- [x] Ran `nix fmt`


george-cosma:
Could you please write some tests that would fail due to this bug but now succeed thanks to your fix(es)? Just so we don't have it happen again.

george-cosma (approved):

nerodesu017 (approved):

PR #128, opened 2025-02-17
Global Store

george-cosma:
### Pull Request Overview

### Testing Strategy

### TODO or Help Wanted

### Formatting

- [ ] Ran `cargo fmt`
- [ ] Ran `cargo check`
- [ ] Ran `cargo build`
- [ ] Ran `cargo doc`
- [ ] Ran `nix fmt`

### Github Issue

### DIFF
--- a/src/execution/store.rs
+++ b/src/execution/store.rs
@@ -1,4 +1,5 @@
 use alloc::collections::btree_map::BTreeMap;
+use alloc::collections::btree_set::BTreeSet;
 use alloc::string::{String, ToString};
 use alloc::vec;
 use alloc::vec::Vec;
@@ -11,7 +12,7 @@ use crate::core::reader::types::element::{ElemItems, ElemMode};
 use crate::core::reader::types::export::ExportDesc;
 use crate::core::reader::types::global::Global;
 use crate::core::reader::types::import::ImportDesc;
-use crate::core::reader::types::{MemType, TableType, ValType};
+use crate::core::reader::types::{FuncType, MemType, TableType, ValType};
 use crate::core::reader::WasmReader;
 use crate::core::sidetable::Sidetable;
 use crate::execution::value::{Ref, Value};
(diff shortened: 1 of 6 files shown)


src/execution/store.rs
```diff
@@ -1,31 +1,481 @@
...
+
+        // TODO: failing is harder since we already modified 'self'. We will circle back to this later.
+
+        // let temp = Vec::new();
+
+        for module_idx in 0..self.modules.len() {
+            for function_idx in 0..self.modules[module_idx].functions.len() {
+                let fn_store_idx = self.modules[module_idx].functions[function_idx];
+                let func: &FuncInst = &self.functions[fn_store_idx];
+                if let FuncInst::Imported(import) = func {
```
cemonem, line 139: global stores might be completely relieving us from Imported/Local enums

cemonem: One more thing is, it might be beneficial to check whether something imported from Y in module X is certainly exported in module Y, by just going over the exports of Y here, so essentially a double for loop going over X's import section in the outer and Y's export section in the inner. Or if there is an issue of 100 exports (i don't think this is worrisome) on the Y's end, simply a hashmap that holds module,name pairs as keys with the store address of the instance

src/execution/store.rs
```diff
@@ -1,31 +1,481 @@
...
+                ty: *ty,
+                locals,
+                code_expr,
+                // TODO figure out where we want our sidetables
+                sidetable: sidetable.clone(),
+                function_type: self.types[*ty].clone(),
+            })
+        });
+
+        let imported_function_inst = self.imports.iter().filter_map(|import| match &import.desc {
```
cemonem, line 265: This part isn't necessary, the imported functions would be already registered to the store when the imported module was instantiated before. It's more logical for a module to have an imports field instead of a store, which only serves purpose during validation for types and checking whether imported entities were instantiated before just before this modules instantiation

src/execution/store.rs
```diff
@@ -1,31 +1,481 @@
...
+
+            let code_expr = wasm_reader
+                .make_span(func.len() - bytes_read)
+                .expect("TODO remove this expect");
+
+            FuncInst::Local(LocalFuncInst {
+                ty: *ty,
+                locals,
+                code_expr,
+                // TODO figure out where we want our sidetables
```
cemonem, line 259: Sidetables would ideally be a field of modules, and would be per-module with along with their source code. In our current implementation we still have them per function, they can stay this way for now

florianhartung:
closed in favor of https://github.com/DLR-FT/wasm-interpreter/pull/131

PR #129, opened 2025-02-18
feat: automatic nix docker environment

eva-cosma:
### Pull Request Overview

This PR adds a small bash script to auto-manage a docker container using `nixos/nix:latest`. This is useful for avoiding direct installation of nix for a more versatile experience (i.e. my laptop's setup is hanging on by a thread, I am afraid to install nix directly and I don't want to deal with virtual machines)

I put these scripts under the `ci-tools` folder, but if you think that it isn't appropriate we can move it elsewhere.

### DIFF
--- a/ci-tools/start_nix.sh
+++ b/ci-tools/start_nix.sh
@@ -1,0 +1,66 @@
+#!/usr/bin/env bash
+
+echo "-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-="
+echo "This script is not actively maintained by the"
+echo "development team. Please run at your own risk."
+echo "Feel free to report any issues on our github,"
+echo "but we offer no guarantees of a fix being"
+echo "implemented."
+echo "-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-="
+
+if ! grep -q WSL /proc/version && grep -q MINGW /proc/version 2>/dev/null; then
+    echo "This script requires WSL if ran on windows for proper path handling with Docker."
+    echo "Please run this script from within WSL instead of Git Bash."
+    exit 1
+fi
+
+
+# Make sure that we are in the `ci-tools` directory
+current_dir=$(basename "$(pwd)")
+if [ "$current_dir" != "ci-tools" ]; then
+    if [ -d "ci-tools" ]; then
+        cd "ci-tools" || exit 1
+    else
+        echo "Error: ci-tools dir not found -- where are you running this from?"
+        exit 1
+    fi
+fi
+
+mkdir -p nix-sharedfs
+docker build -t nix-playground .
+
+# Copy the parent directory contents into the container's workspace. This way, any results do not backspill onto the
+# host which can mess up wsl/windows. We want to copy only the things git would copy (so we don't copy over `.git` or
+# `target` or anything else git wouldn't copy)
+
+temp_tar=$(mktemp)
+cd ..
+git archive --format=tar HEAD > "$temp_tar"
+cd - || exit 1
+
+CONTAINER_NAME="nix-playground-container"
+
+if ! docker container inspect "$CONTAINER_NAME" >/dev/null 2>&1; then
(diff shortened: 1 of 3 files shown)


ci-tools/start_nix.sh
```diff
@@ -0,0 +1,58 @@
...
+    fi
+fi
+
+mkdir -p nix-sharedfs
+docker build -t nix-playground .
+
+# Copy the parent directory contents into the container's workspace. This way, any results do not backspill onto the
+# host which can mess up wsl/windows. We want to copy only the things git would copy (so we don't copy over `.git` or
+# `target` or anything else git wouldn't copy)
+
```
cemonem, line 27: If i was using a devcontainer I would find it more comfortable for the things to change on my end so I'd just bind mount whatever cloned from the repo and start hacking. but if you say its necessary, then alright

ci-tools/start_nix.sh
```diff
@@ -0,0 +1,58 @@
...
+if ! docker container inspect "$CONTAINER_NAME" >/dev/null 2>&1; then
+    echo "Creating new container..."
+
+    docker run -dit \
+        --name "$CONTAINER_NAME" \
+        -v "$(pwd)/nix-sharedfs:/nix-sharedfs" \
+        --workdir /workspace \
+        nix-playground \
+        bash -c "mkdir -p /workspace && cd /workspace && bash"
```
cemonem, line 43: maybe throw somewhere here  `nix develop` or `nix build .#wasm-interpreter ` to download dependencies and build, and docker run with `nix shell` to start from dev environment right away? It's good like this too

cemonem:
I think this is a great idea, would make our project more accessible

ci-tools/start_nix.sh
george-cosma: To this day Windows/wsl no longer recognizes my "wasm-interpreter" folder as a real folder, but rather as a mount of docker because the "result" symlink to the nix store thing confused both systems so much. If I could make it a bind mount, I would, but a shared folder is a much safer bet. Changes made by nix (i.e. nix flake check) can be exported via a git patch file.

This does bring up an issue: exporting git patch files is kinda annoying since you have to re-initialize the git repo. Adding this to the initialization commands could be useful.

dadada:
You can do something like this with Nix:

```nix
packages.devImage = pkgs.dockerTools.buildImage {
  name = "webassembly-interpreter-dev";
  tag = "latest";
  copyToRoot = [ self.devShells.${system}.default ];
  config = {
    Cmd = [ "/entrypoint" ];
    Workdir = "/workdir"
  };
};
```
```command
podman run --mount=type=bind,source=./.,destination=/workspace -w /workspace 6c19e6e64853 ../entrypoint
```

george-cosma:
While the solution looks really neat, it might not work for what I want. If I understood correctly, the docker image is built in the CI and the workspace is that from the CI. So if I have an un-commited local change, it won't be present in the image. Is this right?

There is however a way to satisfy my desire for these scripte to be saved *somewhere* while also being usable on any branch - git worktree. I can create a branch with these scripts, and have it checked out at the same time as the main repo. I can do this on my local fork or in this repo. I don't really care, so it's up to you folks.

wucke13:
> While the solution looks really neat, it might not work for what I want. If I understood correctly, the docker image is built in the CI and the workspace is that from the CI. So if I have an un-commited local change, it won't be present in the image. Is this right?

That is true.

If this works for @george-cosma on windows, I'm fine with it. I have no expertise to provide good Windows support, neither am I able to judge it.

wucke13:
Please declare in the dockerfile, the shell script or in a `ci-tools/README.md` that the windows tooling is officially unsupported and just a best effort.

cemonem (approved):

wucke13:
@cemonem @george-cosma I took the liberty to address my remaining gripes with the PR. I think its good to go, if you agree provide a approval to seal of the deal :smile:

george-cosma:
I approve @wucke13 , but github does not allow me to approve a PR I created

cemonem (approved):

Issue #135, opened 2025-03-07
On the matter of shared mutable access to linear meory for the threading proposal


wucke13:
I drafted a simple implementation of a linear memory which allows shared mutable access without invoking UB (beyond being racy, of course). It does however currently rely on `RwLock`, which is not available in libcore. So we do have to create a `RwLock` before it can be incorporated into the code. Also it still needs some love on the `LittleEndianBytes` trait.

```rust
use std::{cell::UnsafeCell, io::IntoInnerError, sync::RwLock};

///  # Unsafe Note
///
/// Raw pointer access it required, because concurent mutation of the linear memory might happen
/// (consider the threading proposal for WASM, where mutliple WASM threads access the same linear
/// memory at the same time). The inherent race condition results in UB w/r/t the state of the `u8`s
/// in the inner data. However, this is tolerable, e.g. avoiding race conditions on the state of the
/// linear memory can not be the task of the interpreter, but has to be fulfilled by the interpreted
/// bytecode itself.
pub struct LinearMemory {
    inner_data: RwLock<Vec<UnsafeCell<u8>>>,
}

impl LinearMemory {
    // pub const PAGE_SIZE: usize = 1 << 16; // 64 KiB
    pub const PAGE_SIZE: usize = 1 << 12; // 4 KiB for debugging

    /// Create a new, empty [`LinearMemory`]
    pub fn new() -> Self {
        Self {
            inner_data: RwLock::new(Vec::new()),
        }
    }

    /// Create a new, empty [`LinearMemory`]
    pub fn new_with_capacity(pages: i16) -> Self {
        let size_bytes = Self::PAGE_SIZE * pages as usize;
        let mut data = Vec::with_capacity(size_bytes);
        data.resize_with(size_bytes, || UnsafeCell::new(0));

        Self {
            inner_data: RwLock::new(data),
        }
    }

    pub fn grow(&self, pages_to_add: i16) {
        let mut lock_guard = self.inner_data.write().unwrap();
        let prior_length_bytes = lock_guard.len();
        let new_length_bytes = prior_length_bytes + Self::PAGE_SIZE * pages_to_add as usize;
        lock_guard.resize_with(new_length_bytes, || UnsafeCell::new(0));
    }

    pub fn store<T: LittleEndianBytes>(&self, index: i32, value: T) {
        let value_size = std::mem::size_of::<i32>();

        let lock_guard = self.inner_data.read().unwrap();
[...]

PR #136, opened 2025-03-10
Add `RwLock` and concurrency enabled `LinearMemory`

wucke13:
### Pull Request Overview

Adds:

- `RwSpinLock`, a spinning, writer preferring read-write lock implementation
-  `LinearMemory`, a linear memory implementation that enable shared concurrent access

### Testing Strategy

This pull request was tested by its including tests

### TODO or Help 

This pull request still needs a keen eye on the use of ordering for the atomic operations in the `RwSpinLock` implementation.

### Formatting

- [x] Ran `cargo fmt`
- [x] Ran `cargo check`
- [x] Ran `cargo build`
- [x] Ran `cargo doc`
- [x] Ran `nix fmt`

### Github Issue

This pull request continues the ideas laid out in https://github.com/DLR-FT/wasm-interpreter/issues/135

### DIFF
--- a/src/core/little_endian.rs
+++ b/src/core/little_endian.rs
@@ -1,0 +1,37 @@
+//! This module contains the definition and implementation of [`LittleEndianBytes`], a trait to
+//! convert values (such as integers or floats) to bytes in little endian byter order
+
+/// This macro implements the [`LittleEndianBytes`] trait for a provided list of types.
+///
+/// # Assumptions
+///
+/// Each type for which this macro is executed must provide a `from_le_bytes` and `to_le_bytes`
+/// function.
+macro_rules! impl_LittleEndianBytes{
+        [$($type:ty),+] => {
+
+            $(impl LittleEndianBytes<{ ::core::mem::size_of::<$type>() }> for $type {
+                fn from_le_bytes(bytes: [u8; ::core::mem::size_of::<$type>()]) -> Self {
+                    Self::from_le_bytes(bytes)
+                }
+
+                fn to_le_bytes(self) -> [u8; ::core::mem::size_of::<$type>()] {
+                    self.to_le_bytes()
+                }
+            })+
+        }
+    }
+
+/// Convert from and to the little endian byte representation of a value
+///
+/// `N` denotes the number of bytes required for the little endian representation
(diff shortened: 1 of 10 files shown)


george-cosma:
The code looks good overall, but I would like a soundness check w/r/t to atomic ordering. I'm not super familiar with them, but I know of the importance of getting them right. Perhaps we can make this a topic for tomorrow's meeting

Also, for the spinlock, while I have no gripes of setting u32::MAX/2 as a soft-maximum and having u32::MAX as a "writer present flag", it isn't really that rust-y. A rogue `+1` with the `u32::MAX` set can result in an underflow/panic. If we believe we are not going to touch this code in the near future, we can leave this as is, since it looks like a correct implementation. If not, having an inline function to turn the AtomicU32 (after get) into an enum variant that we then check, or the other way around (before set) can reduce the chance of a logic bug appearing.

src/core/spinlock.rs
```diff
@@ -0,0 +1,154 @@
...
+        // this is what makes this lock biased towards preferring writers
+        while self.waiting_writers.load(Ordering::Acquire) != 0 {
+            hint::spin_loop();
+        }
+
+        // get the current number of readers
+        let mut s = self.state.load(Ordering::Relaxed); // ordering by the book
+        loop {
+            // while there are either too many readers or an active writer, wait
+            if s > u32::MAX / 2 {
```
cemonem, line 66: Cem: show the case where this u32::MAX/2, or u32::MAX-1 (anything except u32::MAX ) is really needed

cemonem: I could not find any reason for not putting u32::MAX here, except there are U32::MAX readers, and the U32::MAXth reader updates self.state to u32::MAX which could be percieved as a writer obtaining the lock for itself, though that in itself does not create a problem as far as i see.

wucke13:
@george-cosma @cemonem I tried to apply your feedback, fix the bugs, improve the locking a bit further, now with ordering completely identical to the mara bos book. I also extended the documentation. From my point of view, this PR is good. Please have a look if you find any more gotchas :smile:

nerodesu017 (approved):

cemonem (approved):

PR #142, opened 2025-03-17
feat: integer extend op codes

nerodesu017:
### Pull Request Overview

This pull request adds support for integer extend opcodes

### Testing Strategy

This pull request was tested using the testsuite i32.wast and i64.wast files

### TODO or Help Wanted

N/A

### Formatting

- [x] Ran `cargo fmt`
- [x] Ran `cargo check`
- [x] Ran `cargo build`
- [x] Ran `cargo doc`
- [ ] Ran `nix fmt`

### Github Issue

N/A

### DIFF
--- a/src/execution/interpreter_loop.rs
+++ b/src/execution/interpreter_loop.rs
@@ -2867,6 +2867,94 @@ pub(super) fn run<H: HookSet>(
                     _ => unreachable!(),
                 }
             }
+
+            I32_EXTEND8_S => {
+                let mut v: u32 = stack.pop_value(ValType::NumType(NumType::I32)).into();
+
+                if v | 0xFF != 0xFF {
+                    trace!("Number v ({}) not contained in 8 bits, truncating", v);
+                    v &= 0xFF;
+                }
+
+                let res = if v | 0x7F != 0x7F { v | 0xFFFFFF00 } else { v };
+
+                stack.push_value(res.into());
+
+                trace!("Instruction i32.extend8_s [{}] -> [{}]", v, res);
+            }
+            I32_EXTEND16_S => {
+                let mut v: u32 = stack.pop_value(ValType::NumType(NumType::I32)).into();
+
+                if v | 0xFFFF != 0xFFFF {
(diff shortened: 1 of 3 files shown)


cemonem (dismissed):

src/execution/interpreter_loop.rs
```diff
@@ -2867,6 +2867,94 @@ pub(super) fn run<H: HookSet>(
                     _ => unreachable!(),
                 }
             }
+
+            I32_EXTEND8_S => {
+                let mut v: u32 = stack.pop_value(ValType::NumType(NumType::I32)).into();
+
+                if v | 0xFF != 0xFF {
+                    trace!("Number v ({}) not contained in 8 bits, truncating", v);
```
george-cosma, line 2875: Is this behaviour specified in the spec? The truncating i mean

nerodesu017: To be honest, without this behavior, tests in the official testsuite wouldnt pass. I havent checked the actual implementation, but will do

george-cosma: 1. https://webassembly.github.io/spec/core/appendix/index-instructions.html
2. https://webassembly.github.io/spec/core/exec/numerics.html#op-iextendn-s
3. https://webassembly.github.io/spec/core/exec/numerics.html#op-wrap

It seems that this is the intended behaviour. Can you double check that mod == the same thing you are doing in this case? It should be, but to be safe...

nerodesu017: Indeed have checked with wasmi and wizard engine. What they do is they just recast to the wanted type (i8, i16 or i32) and then sign-extend. Essentially being the same thing as forcefully &= with the desired 2^N and then sign-extending

wucke13 (approved):

src/execution/interpreter_loop.rs
wucke13: Good, then this one is resolved :smile:

wucke13 (approved):

cemonem (approved):

PR #143, opened 2025-03-20
fix: Catch Unwind in direct invoke directive

george-cosma:
Please see:
- #150


florianhartung:
closed in favor of https://github.com/DLR-FT/wasm-interpreter/pull/146

PR #146, opened 2025-03-21
Dev/testsuite rework 2

george-cosma:
Please see:
- #150


nerodesu017 (approved):

george-cosma:
I find it intresting that code coverage went down despite test resulte being the same. Perhaps there is a script/compulation error somewhere when originally it treated it as a simple error instead?

florianhartung:
closed in favor of https://github.com/DLR-FT/wasm-interpreter/pull/147

PR #151, opened 2025-04-01
Nero/dev/global store real assert unlinkable

cemonem:
I checked the reference interpreter and each AssertUnlinkable directive in the .wast files. It seems AssertUnlinkable only checks for two exceptions. Eval.Link and Import.Unknown:
https://github.com/WebAssembly/spec/blob/05949f507908aac3ad2a21661b5c39fa013da950/interpreter/script/run.ml#L472

which are only thrown when:
- there is no such import
- there is a type mismatch

https://github.com/WebAssembly/spec/blob/05949f507908aac3ad2a21661b5c39fa013da950/interpreter/script/import.ml#L17
https://github.com/WebAssembly/spec/blob/05949f507908aac3ad2a21661b5c39fa013da950/interpreter/exec/eval.ml#L732
https://github.com/WebAssembly/spec/blob/05949f507908aac3ad2a21661b5c39fa013da950/interpreter/exec/eval.ml#L788

These checks do not alter the store. Therefore no need for CleanupStore :)
Though I think even though it sounds prepostrous, we might need to be able to copy the entire store. For tests, at least, since if there is an erroneous test altering store state we need to revert somehow.

Finally in the future, for initialization order, we might want to refer to this:
https://webassembly.github.io/spec/core/exec/modules.html#instantiation

Item 11 seems small but it actually expands to https://webassembly.github.io/spec/core/exec/modules.html#alloc-module

### DIFF
--- a/src/execution/execution_info.rs
+++ b/src/execution/execution_info.rs
@@ -3,6 +3,7 @@ use alloc::vec::Vec;
 
 use crate::core::reader::types::export::Export;
 use crate::core::reader::types::FuncType;
+use crate::core::sidetable::Sidetable;
 // use crate::core::reader::types::FuncType;
 // use crate::execution::Store;
 
@@ -12,6 +13,8 @@ use crate::core::reader::types::FuncType;
 pub struct ExecutionInfo<'r> {
     pub name: String,
     pub wasm_bytecode: &'r [u8],
+    //TODO turn this into ref
+    pub sidetable: Sidetable,
 
     pub functions: Vec<usize>,
     pub functions_offset: usize,
--- a/src/execution/execution_info.rs
+++ b/src/execution/execution_info.rs
@@ -4,7 +4,6 @@ use alloc::vec::Vec;
 use crate::core::reader::types::export::Export;
 use crate::core::reader::types::FuncType;
 // use crate::core::reader::types::FuncType;
-use crate::core::reader::WasmReader;
 // use crate::execution::Store;
 
 /// ExecutionInfo is a compilation of relevant information needed by the [interpreter loop](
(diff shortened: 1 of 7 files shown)


nerodesu017:
Looks good to me, but I still have a few questions:
1. How many tests from the testsuite pass? On my branch 26030 were passing, 228 failing, with 5 compilation errors
2. How would you suggest we make the store cloneable, then? My idea for making sure the store's state doesn't mutate was to revert back using (and possible expanding) the cleanup-related struct. We still need to mutate state and have some kind of, what in my opinion is, an invalid state (but still functional), because of the way the spec for v2 is

cemonem:
> Looks good to me, but I still have a few questions:
> 
>     1. How many tests from the testsuite pass? On my branch 26030 were passing, 228 failing, with 5 compilation errors
> 
>     2. How would you suggest we make the store cloneable, then? My idea for making sure the store's state doesn't mutate was to revert back using (and possible expanding) the cleanup-related struct. We still need to mutate state and have some kind of, what in my opinion is, an invalid state (but still functional), because of the way the spec for v2 is

1. I compared stdout and stderr of both branches with cargo test --test wasm_spec_testsuite -- --nocapture , they were the same
2. The problem goes beyond that with other tests, where the memory is modified in some memory instruction test. Then cloning is the only option. Maybe there is a way to implement Clone trait locally within wasm_spec_testsuite?

wucke13:
W/r/t to the store cleanup (without having seen the code): I think, in production it is very unlikely that we link a module into an existing store, only then to find out that linking failed, subsequently necessitating the failed linking to be undone. I'd rather have the run-time cost of clone for these cases than the implementation cost of complicated book-keeping to enable transactional linking. If of course we get transactional linking for free (as it seems Cem claims?), than of course lets take it.

florianhartung (approved):

Issue #155, opened 2025-04-09
Rename ExecutionInfo to ModuleInst


florianhartung:
and also add a link to the ModuleInst Wasm spec page

cemonem:
fixed in #167

PR #161, opened 2025-05-16
Dev/cemonem/global store refactor

### LINKED ISSUE #153: `ExecutionInfo` became is pregnant with a `ModuleInstance`
- `ExecutionInfo` currently contains a lot of things that belong into a `ModuleInstance`
- `ModuleInstance` does not yet exist
-  `ExecutionInfo` is part of the store
-  `ExecutionInfo` contains the stateful `WasmReader`, but reader state should not be part of the store

To Address this:

- We should rename `ExecutionInfo` -> `ModuleInstance`
- We should remove the `WasmReader` from the `ModuleInstance`
- We should add a new `ExecutionInfo` which can contain both the `Stack`, the `WasmReader` and the current module index
- A `&mut ExecutionInfo` should be passed to the interpreter loop

cemonem:
wip for #153  #155 and spec honoring initialization order

This pull request addresses the folowing issues:

- #153 ExecutionInfo is replaced with ModuleInst struct. This struct tries to match to the spec except the inclusion of the source bytecode and sidetable.
- The initialization store.add_module() is rewritten to match the spec as much as possible. The initialization step  orders are honored wrt spec. Table, Memory active/declarative initialization implementations are also included
- Instance allocation methods defined in the spec are implemented.
- call to start method is moved from RuntimeInstance to initialization.
- Implementation of ref.func instruction and ref values are fixed. The previous refs held func_idx's, which would be erroneous if ref a was to be pushed from module X and was call_indirect'ed from an imported function from module Y.

I suggest reviewing store.rs::add_module side by side with here: https://webassembly.github.io/spec/core/exec/modules.html#instantiation

### Pull Request Overview

### Testing Strategy

### TODO or Help Wanted

- ModuleInst field names should be the same as spec, but that creates a larger diff which makes this big change already hard to review.
- validation_info is extremely similar to Ast.Module in the spec, but is not the same thing exactly. Nevertheless, add_module uses it as Ast.Module, which might be problematic in the future.
- FuncInst mismatches the definition in spec. The spec definition also includes a back pointer to the owning ModuleInst, which needs discussion in our codebase.
- Name changes: if an index makes sense within the store, it should be _addr. If it makes sense locally within the module, it should be _idx.
- Error reporting strategy during Initialization needs to be determined.
-
- FunctionRef and RuntimeInstance structs and their impls might have to be refactored/removed. There is code duplication between these and Store.
- instruction loop run() signature and the execution stack structure and methods might have to be refactored/reviewed. The nature of these structs makes some spec honoring implementations harder.
- Store holds references to module instances and a kv store to module names. Curiously the spec definition of the store does not include it, there is no such thing as a module_addr/module_idx in the spec. Similarly, our initialization includes linking related activities at the beginning, which is not mentioned in the spec and instead external values are supplied to module_instantiate method seperately.

### Formatting

- [ ] Ran `cargo fmt`
- [ ] Ran `cargo check`
- [ ] Ran `cargo build`
- [ ] Ran `cargo doc`
- [ ] Ran `nix fmt`

### Github Issue

This pull request closes #153 #155


src/core/error.rs
cemonem, line 1: Our errors have become too complicated. I think we need to start from a clean state where we know what error to throw when, and adhere to that by heart. Otherwise we are adding errors here everytime, causing overlapping or unused errors. One thing this definitely creates a problem is the wasm_spec_testsuite, in which we have to translate some of our errors, and there might be cases when an error thrown from us corresponds to two different errors in the original spec. I suggest:
1. we nuke this file and start with only the errors thrown from the spec ocaml code
2. with the initial set of errors, create a 1-1 mapping to spec code errors (for wasm_spec_testsuite checks)
3. slowly replace these errors with our custom ones, carefully turning that 1-1 mapping into an onto mapping, being careful not to map to two different spec errors.

src/core/reader/types/element.rs
cemonem, line 1: Elements have a special property that some of the initializer expressions are not in the form of instructions, forcing us to divert from the spec when dealing with them.

src/core/reader/types/mod.rs
```diff
@@ -457,3 +458,14 @@ impl WasmReadable for MemType {
         }
     }
 }
+
```
cemonem, line 461: I hadn't noticed there was a subtyping relation between imports and exports. This might be a thing that should be attacked after this hopefully merges.

src/execution/function_ref.rs
cemonem, line 1: I am highly in favor of  future careful refactoring/renaming/removing FunctionRef to something that corresponds to it in the spec. But as a start, when newtyping *_addrs, the functionality of this can simply go there.

src/execution/interpreter_loop.rs
cemonem, line 1: In this file:
- I made minimal enough changes in names for sanity for *_addrs and *_idx's for representing indices that make sense in the store or in the module respectively.
- I fixed ref.func and corresponding call_indirect.
- I encapsulated table_init, elem_drop, memory_init, data_drop for initialization.

src/execution/mod.rs
cemonem, line 1: In this file:
- 
- I moved the call to start function during initialization, to add_module. It should not be here, otherwise we risk multiple calls to start function, which the spec tells us to call once.
- I made minimal enough changes in names for sanity for *_addrs and *_idx's for representing indices that make sense in the store or in the module respectively.
- I removed FunctionRef::verify_func() and all of the sort, because it is not on the spec and it can be easily circumvented. With the global store, its worth is less than maintaining it.

src/execution/store.rs
cemonem, line 1: This file has the main changes this PR targets.
- It tries to implement add_module (which I am in the favor of naming instantiate) in the wasm instance alllocation/initialization order the spec dictates.
- It also implements allocation methods specified in the spec.

cemonem:
closed by https://github.com/DLR-FT/wasm-interpreter/pull/167

PR #181, opened 2025-05-28
Dev/wucke13 wasmspec fix stack exhaustion failure

wucke13:
### Pull Request Overview

Implements stack exhaustion checks and tester assert exhaustion directives.

### Checks

- Using Nix
  - [x] Ran `nix fmt`
- Using Rust tooling
  - [x] Ran `cargo fmt`
  - [x] Ran `cargo test`
  - [x] Ran `cargo check`
  - [x] Ran `cargo build`
  - [x] Ran `cargo doc`

### DIFF
--- a/src/core/error.rs
+++ b/src/core/error.rs
@@ -17,7 +17,6 @@ pub enum RuntimeError {
     DivideBy0,
     UnrepresentableResult,
     FunctionNotFound,
-    StackSmash,
     // https://github.com/wasmi-labs/wasmi/blob/37d1449524a322817c55026eb21eb97dd693b9ce/crates/core/src/trap.rs#L265C5-L265C27
     BadConversionToInteger,
     MemoryAccessOutOfBounds,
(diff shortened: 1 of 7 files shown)


florianhartung:
About the newly introduced partial coverage: I think we should use `.unwrap()` or `.expect("")` in those cases, where stack exhaustion is impossible.
For example consider the `i32.add` instruction, which first pops two values and then pushes one. Pushing a new value should *never* exhaust the stack, making the use of unwrap justified.


This way we can reduce the partially covered lines, which are impossible to reach anyway.
Note: To my knowledge the statement `xxxx?` is partially covered and `xxxx.unwrap()` fully covered if the error path is not taken during execution.

wucke13:
I do not disagree in general, but there is a point to be made: what you say is true until when a stack limit changes at runtime. Currently they are constant so all fine, but...

nerodesu017 (approved):
Looks reaaaaaaally good to me, glad to see it be 100% compliant with the test suite 🥰
S/O to you all!

PR #185, opened 2025-06-05
fix: polymorphic select validation

cemonem:
### Pull Request Overview

This pull request fixes the polymorphic select instruction's validation, the previous one was too tight in unifying the type variable on top of the stack. Some names were also changed to better match with the spec.

The unreached-valid test actually validated the test containing a ref.is_null after a polymorphic select, but failed to validate a subsequent module #186 (that's why the test results seem not to have changed) The PR originally aimed to fix that issue as well but that one turned out to require some heavy changes, so I am excluding it for now.

### Checks

- Using Nix
  - [x] Ran `nix fmt`
- Using Rust tooling
  - [x] Ran `cargo fmt`
  - [x] Ran `cargo test`
  - [x] Ran `cargo check`
  - [x] Ran `cargo build`
  - [x] Ran `cargo doc`

### DIFF
--- a/src/validation/validation_stack.rs
+++ b/src/validation/validation_stack.rs
@@ -4,7 +4,7 @@ use alloc::vec::Vec;
 
 use crate::{
     core::reader::types::{FuncType, ResultType},
-    Error, RefType, ValType,
+    Error, NumType, RefType, ValType,
 };
 
 #[derive(Debug, PartialEq, Eq)]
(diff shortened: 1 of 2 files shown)


src/validation/validation_stack.rs
```diff
@@ -306,73 +293,54 @@ impl ValidationStack {
...
-            ValidationStackEntry::NumOrVecType => Ok(self.clone()),
-            ValidationStackEntry::UnspecifiedValTypes => Ok(ValidationStackEntry::NumOrVecType),
-            _ => Err(Error::TypeUnificationMismatch),
+    /// convenience method that returns `Some(other)` if `self.unifies_to(other)` is true and `None` otherwise
+    fn unify(&self, other: &ValidationStackEntry) -> Option<Self> {
+        if self.unifies_to(other) {
+            Some(other.clone())
+        } else {
+            None
         }
```
florianhartung, line 344:
```suggestion
fn unify(&self, other: &ValidationStackEntry) -> Option<Self> {
        self.unifies_to(other).then(|| other.clone())
```

florianhartung (changes requested):

florianhartung (approved):

PR #197, opened 2025-06-23
fix: remove non const instrs from validation and const evaluation

cemonem:
### Pull Request Overview

this pr removes non const instrs from const validation and const evaluation which creates pain in coverage tests.

### Checks

- Using Nix
  - [x] Ran `nix fmt`
- Using Rust tooling
  - [x] Ran `cargo fmt`
  - [x] Ran `cargo test`
  - [x] Ran `cargo check`
  - [x] Ran `cargo build`
  - [x] Ran `cargo doc`

### DIFF
--- a/src/execution/const_interpreter_loop.rs
+++ b/src/execution/const_interpreter_loop.rs
@@ -6,7 +6,7 @@ use crate::{
     },
     value::{self, FuncAddr, Ref},
     value_stack::Stack,
-    ModuleInst, NumType, RefType, Store, ValType, Value,
+    ModuleInst, RefType, Store, Value,
 };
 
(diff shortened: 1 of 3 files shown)


florianhartung (approved):

florianhartung:
@wucke13 You are correct, specifically for numeric types only the `t.const c` instruction seems to be allowed.

https://webassembly.github.io/spec/core/valid/instructions.html#constant-expressions

cemonem:
> @cemonem just to confirm: these instructions are infact not allowed in const expressions and thus could never have been executed?

yes :)

wucke13 (approved):

PR #208, opened 2025-07-10
Whitepaper: small fixes 

puzzlewolf:
### Pull Request Overview

- bump the dependencies that throw warnings with Typst 0.13.0
- add `strictdoc` to the devshell
- fix the `whitepaper-watch` command by correcting the path

### Checks

- Using Nix
  - [x] Ran `nix fmt`
- Using Rust tooling
  - [ ] Ran `cargo fmt`
  - [ ] Ran `cargo test`
  - [ ] Ran `cargo check`
  - [ ] Ran `cargo build`
  - [ ] Ran `cargo doc`


flake.nix
```diff
@@ -119,6 +119,8 @@
                   cargo-watch
                   wabt
 
+                  strictdoc
+
                   # utilities
```
wucke13, line 124: Is this necessary? I would assume via inputsFrom we already have this

puzzlewolf: Doesn't seem so. The `requirements-*` commands don't work with `strictdoc: command not found`.

wucke13: My mistake. `self.packages.${system}.requirements` is missing from `packagesFrom` in the devshell. Please add it above `self.packages.${system}.whitepaper` to maintain alphanumerical sorting.

PR #209, opened 2025-07-11
fix: clippy lints from future

wucke13:
### Pull Request Overview
Our CI environment picked up some clippy lints, which the clippy provided via Nix does not yet know about. This commit fixes the lints (all about the use of the `format!` macro).

### TODO or Help Wanted

This pull request still needs to verify that fixing this does not break our MSRV

### Checks

- Using Nix
  - [x] Ran `nix fmt`
- Using Rust tooling
  - [ ] Ran `cargo fmt`
  - [ ] Ran `cargo test`
  - [ ] Ran `cargo check`
  - [ ] Ran `cargo build`
  - [ ] Ran `cargo doc`

### DIFF
--- a/src/core/error.rs
+++ b/src/core/error.rs
@@ -135,8 +135,7 @@ impl Display for Error {
                 "A value was expected in the WASM binary but the end of file was reached instead",
             ),
             Error::InvalidSection(section, reason) => f.write_fmt(format_args!(
-                "Section '{:?}' invalid! Reason: {}",
-                section, reason
+                "Section '{section:?}' invalid! Reason: {reason}"
             )),
             Error::InvalidSectionType(ty) => f.write_fmt(format_args!(
                 "An invalid section type id was found in a section header: {ty}"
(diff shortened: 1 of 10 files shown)


puzzlewolf:
Yeah, applying the lint works for now :)

The underlying issue is that the No-Nix-CI probably pulled a new version of the base image which has a different version of `clippy` (1.88 I think) than the nix-based CI (1.86 at the moment in from current stable `nixpgks`. This kind of issue is bound to happen again. 

Do we need the No-Nix-CI? If yes, I think we should make it use the same Rust version as we develop on. Or maybe use the MSRV, as that's already stated explicitly. There are Debian or Alpine based [docker images](https://hub.docker.com/_/rust) for specific Rust versions, maybe those are useful?

wucke13:
Strictly speaking, we do not need it (the No-NIx CI). However, I'd say it is actually valuable to have something which is not declarative, just pulling in the newest version of the tool-chain, to have a prospect on problems which in the future _will be_ coming to us.

> This kind of issue is bound to happen again.

So I consider this a feature. In the end, other might (hopefully?!) also use this code, but not our CI setup/Nix cruft, and I want them to have a smooth experience. That exact experience is what we take a glance on when running "whatever happens to be the current stable toolchain" in CI.

puzzlewolf:
I agree that running CI on whatever is currently stable provides useful information. But it's not nice to have PRs fail for external reasons. So, maybe don't fail the pipeline if No-Nix fails?

florianhartung (approved):

PR #215, opened 2025-07-24
fix: func_addrs in stack instead of module_addr, func_idx pairs

cemonem:
### Pull Request Overview

func_idx, module_addr pairs are basically the same as func_addrs, and they make no sense when it is a host function. Also, during invocation, a hacky linear search had to be made for retrieving func_idx of a function from its func_addr. This fixes this.

### TODO or Help Wanted

- The current work in this branch is a preparation for moving the stack initialization logic within store.invoke_* into interpreter_loop::run.

### Checks

- Using Nix
  - [x] Ran `nix fmt`
- Using Rust tooling
  - [x] Ran `cargo fmt`
  - [x] Ran `cargo test`
  - [x] Ran `cargo check`
  - [x] Ran `cargo build`
  - [x] Ran `cargo doc`

### DIFF
--- a/src/execution/interpreter_loop.rs
+++ b/src/execution/interpreter_loop.rs
@@ -45,12 +45,11 @@ pub(super) fn run<H: HookSet>(
     let mut current_func_addr = func_addr;
     let func_inst = &store.functions[current_func_addr];
     let mut current_module_idx = func_inst.module_addr;
-    let current_module_idx = &mut current_module_idx;
 
     // Start reading the function's instructions
-    let wasm = &mut WasmReader::new(store.modules[*current_module_idx].wasm_bytecode);
+    let wasm = &mut WasmReader::new(store.modules[current_module_idx].wasm_bytecode);
 
-    let mut current_sidetable: &Sidetable = &store.modules[*current_module_idx].sidetable;
+    let mut current_sidetable: &Sidetable = &store.modules[current_module_idx].sidetable;
     let mut stp = func_inst.stp;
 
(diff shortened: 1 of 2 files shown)


wucke13 (dismissed):

florianhartung (approved):

src/execution/interpreter_loop.rs
```diff
@@ -37,15 +37,15 @@ use super::store::Store;
...
     store: &mut Store,
 ) -> Result<(), RuntimeError> {
-    let global_func_idx =
-        store.modules[*current_module_idx].func_addrs[stack.current_stackframe().func_idx];
-
-    let func_inst = store.functions.get(global_func_idx).unwrap_validated();
+    let mut current_func_addr = func_addr;
+    let func_inst = &store.functions[current_func_addr];
+    let mut current_module_idx = func_inst.module_addr;
+    let current_module_idx = &mut current_module_idx;
```
florianhartung, line 48: I don't think we need this duplicate variable definition. Also the functions at the very bottom of this file should not immutable references to the current module idx (e.g. `current_module_idx: &usize`) as arguments, but rather take the current module idx directly.

florianhartung (changes requested):

src/execution/interpreter_loop.rs
cemonem: I agree. Only did it to not to increase the diff size, since it's being used everywhere, glad to change it once this gets approved :)

florianhartung (approved):

PR #216, opened 2025-07-24
refactor: move duplicated stack initialization to interpreter loop

cemonem:
### Pull Request Overview

I based this on #215 for now, when it merges, i'll make this draft ready for merge.

This refactor moves execution stack initialization/usage into interpreter loop and removes duplicated code in `store.invoke_*` methods.

A discussion point is, it also adds a validation check that panics in an `unreachable_validated!` way if there is a type mismatch into the last part of the interpreter loop. This check was in the `store.invoke_*` methods already. I'm not sure it should exist at all.

### TODO or Help Wanted
- The parts where very first stack initialization/function call/function return takes place in the interpreter loop have some duplication we can tend to. I want to tackle this next.

- should the type checking at the end of the `interpreter_loop::run` be removed?

- Now that `invoke_*` are thinner, they are easier to discuss about. I have the following suggestions:

1. Type checking of (input or output) values to an invoke should take place within `RuntimeInstance::invoke_*` and it should throw an `Err`. No type checking that would throw an `Err` should take place within the `store.invoke_*` methods, these methods should assume their inputs are validated (which implies the postcondition of these `invoke_*` methods are the validity of the values they return with output type of the function, here even `RuntimeInstance` doesn't have to check post` interpreter_loop::run()`)
2. The store should have a single invoke method that does no type checking with a signature along the lines
```
fn invoke(&mut self, func_addr: usize, params: Vec<Value>) -> Result<Vec<Value>, Trap)
```
similar to `invoke_dynamic_unchecked_return_ty` we have now. This method can even reduce to a single `interpreter_loop::run` call, but it is hard to tell yet.

3. `RuntimeInstance` should have two methods, one which the input/output types of the function are supplied by hand as generics (similar to our `invoke` -- lets call it `invoke_typed`), and one that the user doesn't supply any input/output type at all (similar to `invoke_dynamic_unchecked_return_ty`, let's call this one `invoke_untyped`). These methods both internally call `store.invoke`, but with the following pre-checks.

3a. `invoke_typed` checks whether types supplied with generics match with the function's input and output types. It is impossible to supply values with incorrect input types if this check goes through since rust type checker checks it for us already.
3b. `invoke_untyped` checks whether the input values match the input signature of the function.

None of these do any checks after `store.invoke` is called in an `Err` throwing way, since wasm type system guarantees we would get the correct output types.

Q: What if the user want to supply types in runtime because they are unsure of the wasm types of the things they deal with during runtime, similar to what our `invoke_dynamic` does?
A: The user can do this manually using `store.func_type` method here https://webassembly.github.io/spec/core/appendix/embedding.html#functions . There is no point in embedding it within another `invoke` variant. Our `invoke_dynamic` is seldomly used in our codebase, and these uses can be easily replaced, for example.
Q: Ok, but wouldn't that also make` invoke_typed` redundant?
A: No, since here the user supplies the types in generics and can enjoy the type system of rust when they write programs. The advantage is too big to accept an `invoke` variant here.

4. We should purge `RuntimeInstance/Store::invoke_dynamic`, and rename `invoke_dynamic_unchecked_return_ty` as `invoke_dynamic` or something along the lines. The store should only have this method (maybe with a shorter name like `invoke`, there is no need for the typed variant here)

### Checks

- Using Nix
  - [x] Ran `nix fmt`
- Using Rust tooling
  - [x] Ran `cargo fmt`
  - [x] Ran `cargo test`
  - [x] Ran `cargo check`
  - [x] Ran `cargo build`
  - [x] Ran `cargo doc`

### DIFF
--- a/src/execution/interpreter_loop.rs
+++ b/src/execution/interpreter_loop.rs
@@ -38,12 +38,20 @@ use super::store::Store;
 /// Interprets a functions. Parameters and return values are passed on the stack.
 pub(super) fn run<H: HookSet>(
     func_addr: usize,
-    stack: &mut Stack,
+    params: Vec<Value>,
     mut hooks: H,
     store: &mut Store,
-) -> Result<(), RuntimeError> {
+) -> Result<Vec<Value>, RuntimeError> {
     let mut current_func_addr = func_addr;
     let func_inst = &store.functions[current_func_addr];
+
+    let mut stack = Stack::new();
+    let locals = Locals::new(params.into_iter(), func_inst.locals.iter().cloned());
+
+    // setting `usize::MAX` as return address for the outermost function ensures that we
+    // observably fail upon errornoeusly continuing execution after that function returns.
+    stack.push_stackframe(usize::MAX, &func_inst.ty(), locals, usize::MAX, usize::MAX)?;
+
     let mut current_module_idx = func_inst.module_addr;
 
     // Start reading the function's instructions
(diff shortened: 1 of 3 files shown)


cemonem:
> I am a bit worried, how moving the stack initialization into the interpreter loop interacts with the planned resumable execution. I fear that the interpreter loop becomes more complex and harder to understand and maintain. IIRC originally the plan was to have the interpreter loop just as the main loop for interpretation (relatively easy to understand).
> 
> My main question is: Why should we move stack initialization into the interpreter loop, when we can also do it in `Store::invoke` (where it would make more sense IMO) and just make `Store::invoke_typed` call `Store::invoke`?

Code very similar to what i moved occurs during function calls and returns. I think its better to have them at least in the same place since when we make changes, coupled implementation stays in the same file. But this might be too opinionated too as you say. Fine by me with both cases.

PR #219, opened 2025-07-29
refactor: replace invoke variants with invoke and invoke_typed

cemonem:
### Pull Request Overview

This PR follows suggestions / reviews in #216 .

~Currently there is no functional change.~ This refactor uncovered that start function type validation was missing. It also fixes it.

Unnecessary return type checking at the end of `invoke` was removed, since wasm validation guarantees that already.

Removes `Store/RuntimeInstance::invoke_dynamic`.
Removes `Store::invoke`.
Renames `RuntimeInstance/Store::invoke_dynamic_unchecked_return_ty` as `Store::invoke`.
Renames `RuntimeInstance::invoke` as `RuntimeInstance::invoke_typed`.

### TODO or Help Wanted

- move input type checking logic to `RuntimeInstance`, since store methods are to assume their inputs are validated.

### Checks

- Using Nix
  - [x] Ran `nix fmt`
- Using Rust tooling
  - [x] Ran `cargo fmt`
  - [x] Ran `cargo test`
  - [x] Ran `cargo check`
  - [x] Ran `cargo build`
  - [x] Ran `cargo doc`

### DIFF
--- a/src/execution/value_stack.rs
+++ b/src/execution/value_stack.rs
@@ -35,6 +35,10 @@ impl Stack {
         Self::default()
     }
 
+    pub(super) fn into_values(self) -> Vec<Value> {
+        self.values
+    }
+
     pub fn drop_value(&mut self) {
         // If there is at least one stack frame, we shall not pop values past the current
         // stackframe. However, there is one legitimate reason to pop when there is **no** current
--- a/crates/benchmark/benches/general_purpose.rs
+++ b/crates/benchmark/benches/general_purpose.rs
@@ -102,7 +102,7 @@ macro_rules! bench_wasm {
                 let bid = BenchmarkId::new("our", n);
                 group.bench_with_input(bid, &n, |b, &s| {
                     b.iter(|| {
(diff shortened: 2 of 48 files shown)


src/execution/value_stack.rs
```diff
@@ -30,6 +30,17 @@ pub(crate) struct Stack {
...
+impl IntoIterator for Stack {
+    type Item = Value;
+
+    type IntoIter = alloc::vec::IntoIter<Self::Item>;
+
+    fn into_iter(self) -> Self::IntoIter {
+        self.values.into_iter()
+    }
+}
+
```
florianhartung, line 43: Stacks themselves are not really meant to be iterated over (this is where confusion regarding the order may arise).
Thus I'd suggest that `Stack` shouldn't implement `IntoIterator`, but instead provide a method to be converted to its raw contents (a `Vec`) with a method `Stack::into_inner`.

For our use case both solutions are the same, but I hope this might make the code a bit more intuitive to read.

florianhartung (approved):

PR #223, opened 2025-07-31
Dev/cemonem/host function calls

cemonem:
### Checks

- Using Nix
  - [ ] Ran `nix fmt`
- Using Rust tooling
  - [ ] Ran `cargo fmt`
  - [ ] Ran `cargo test`
  - [ ] Ran `cargo check`
  - [ ] Ran `cargo build`
  - [ ] Ran `cargo doc`


src/execution/interpreter_loop.rs
```diff
@@ -27,33 +28,42 @@ use crate::{
...
     func_addr: usize,
     stack: &mut Stack,
     mut hooks: H,
     store: &mut Store,
 ) -> Result<(), RuntimeError> {
     let mut current_func_addr = func_addr;
     let func_inst = &store.functions[current_func_addr];
-    let mut current_module_idx = func_inst.module_addr;
+    let FuncClosure::WasmFunc(wasm_func_inst) = &func_inst.closure else {
+        panic!("attempted to call non-wasm native function from run_wasm_func")
```
wucke13, line 50: Do we have a better macro to announce that is is an impossible situation?!

src/execution/interpreter_loop.rs
```diff
@@ -27,33 +28,42 @@ use crate::{
...
 
 #[cfg(feature = "hooks")]
 use crate::execution::hooks::HookSet;
 
 use super::store::Store;
 
-/// Interprets a functions. Parameters and return values are passed on the stack.
-pub(super) fn run<H: HookSet>(
+/// Interprets wasm native functions. Parameters and return values are passed on the stack.
+pub(super) fn run_wasm_func<H: HookSet>(
```
wucke13, line 41: Let's not rename this function.

src/execution/store.rs
```diff
@@ -527,46 +546,81 @@ impl<'b> Store<'b> {
...
+            }
+            FuncClosure::WasmFunc(wasm_func_inst) => {
+                // Prepare a new stack with the locals for the entry function
+                let mut stack = Stack::new();
+                let locals = Locals::new(params.into_iter(), wasm_func_inst.locals.iter().cloned());
+
+                stack.push_stackframe(usize::MAX, &func_ty, locals, usize::MAX, usize::MAX)?;
+
+                // Run the interpreter
+                run_wasm_func(
```
wucke13, line 577: use the module path: `interpreter_loop::run` instead of renaming the function

src/execution/store.rs
```diff
@@ -527,46 +546,81 @@ impl<'b> Store<'b> {
...
 
 #[derive(Debug)]
 // TODO does not match the spec FuncInst
-
 pub struct FuncInst {
+    pub function_type: FuncType,
+    pub closure: FuncClosure,
+}
+
+#[derive(Debug)]
```
wucke13, line 599: Put a docstring to explain why this is `FuncClosure`

src/execution/store.rs
```diff
@@ -527,46 +546,81 @@ impl<'b> Store<'b> {
...
+pub enum FuncClosure {
```
florianhartung, line 600: I'd suggest we rename this to `FuncInst`, remove the old `FuncInst` and put the function_type into both enum variants.
Then we can have a method `type` (or `ty` because Rust) on the enum returning the function type regardless of what enum variant is present.

wucke13 (changes requested):

florianhartung (changes requested):

cemonem:
I'll tidy up the commit history once this gets approved.

src/execution/interpreter_loop.rs
florianhartung: Please add a message to the new `unreachable!` call here.

florianhartung:
LGTM! I will approve once this is ready.

src/execution/interpreter_loop.rs
florianhartung: It can be as as simple as
```rs
unreachable!("the interpreter loop shall only be executed with Wasm functions");
```

florianhartung (approved):

PR #227, opened 2025-08-05
fix: make temporary clone for tests from local

cemonem:
### Checks

- Using Nix
  - [ ] Ran `nix fmt`
- Using Rust tooling
  - [ ] Ran `cargo fmt`
  - [ ] Ran `cargo test`
  - [ ] Ran `cargo check`
  - [ ] Ran `cargo build`
  - [ ] Ran `cargo doc`

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -181,16 +181,19 @@
                     name = "generate-testsuite-report";
                     # TODO maybe accept the name of the target branch as an argument and use it instead of `main`
                     command = ''
-                      cd $PRJ_ROOT
+                      cd -- "$PRJ_ROOT"
+
                       TESTSUITE_SAVE=1 cargo test -- spec_tests --show-output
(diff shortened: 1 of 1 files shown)


florianhartung:
What does this PR change?
As far as I see, it only replaces `$(git config --get remote.origin.url)` with `file://$PRJ_ROOT`, while IMO the former is preferred because it is independent of Nix.
Also the top-level parentheses are removed, although I'm not sure why we had them in the first place.

cemonem:
Yeah, only the parantheses are removed and the clone is made from the local to avoid having to clone it from remote, the benchmark commands were implemented with this strategy, that's why i wanted to have this that way too.
The parantheses were a relic from me being afraid that these scripts might be sourced instead of being executed, which is not the case.

wucke13 (approved):

PR #235, opened 2025-08-19
wip: fix ci

puzzlewolf:
### Checks

- Using Nix
  - [ ] Ran `nix fmt`
- Using Rust tooling
  - [ ] Ran `cargo fmt`
  - [ ] Ran `cargo test`
  - [ ] Ran `cargo check`
  - [ ] Ran `cargo build`
  - [ ] Ran `cargo doc`


wucke13:
We do not understand why this occurred, but we have a functional workaround on main now. Hopefully this won't bite in the future.

PR #256, opened 2025-09-26
Make host functions fallible

florianhartung:
~~These changes make it so that host functions return `Result<Vec<Value>, HostError>` instead of `Vec<Value>`. Here `HostError` is a generic type. This allows host functions to return errors and make the interpreter stop execution.~~

~~In the case where the user requires all host functions to be infallible by design, the type [`core::convert::Infallible`](https://doc.rust-lang.org/core/convert/enum.Infallible.html) can be used as the generic `HostError` like so: `Result<Vec<Value>, Infallible>`~~

These changes make host functions return `Result<..., HaltExecutionError>`, where `HaltExecutionError` is a ZST-struct defined by our library. If a host function returns this error, we halt execution and return `RuntimeError::HostFunctionHaltedExecution` back to the original caller.

This allows the user to implement their own error handling by storing an error in the user data `T` and halting execution. At the same time we never have to handle the error ourselves, which makes this a clean solution for providing the user an error handling mechanism without generics everywhere.

Original design: https://github.com/DLR-FT/wasm-interpreter/pull/256#issuecomment-3414642144

### TODO or Help Wanted

~~I've implemented the basic structure for supporting generic host function errors without polluting the entire codebase with generics. If you @wucke @cemonem are fine with this, I will go ahead and update all the tests to work with this solution.~~ (tests are updated now)

### Checks

- Using Nix
  - [x] Ran `nix fmt`
- Using Rust tooling
  - [ ] Ran `cargo fmt`
  - [ ] Ran `cargo test`
  - [ ] Ran `cargo check`
  - [ ] Ran `cargo build`
  - [ ] Ran `cargo doc`

### DIFF
--- a/src/execution/error.rs
+++ b/src/execution/error.rs
@@ -11,6 +11,8 @@ pub enum RuntimeError {
     HostFunctionSignatureMismatch,
     WriteOnImmutableGlobal,
     GlobalTypeMismatch,
+    /// A host function returned the [`HaltExecutionError`](crate::HaltExecutionError), which caused execution to be halted.
+    HostFunctionHaltedExecution,
 
     // Are all of these instantiation variants? Add a new `InstantiationError` enum?
     InvalidImportType,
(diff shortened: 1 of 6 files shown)


cemonem (dismissed):

florianhartung:
Personally, I think using an `Option<...>` might not be very intuitive for the user.
Therefore, suggest using a `Result<..., StopExecution>`, where `StopExecution` is defined by us as a ZST:
```rs
pub struct StopExecution;
```

(there might be a better name for this struct)

florianhartung:
I've moved the changes that required a newer MSRV and made CI fail over to https://github.com/DLR-FT/wasm-interpreter/pull/274.

This PR is ready now.

PR #263, opened 2025-10-06
feat: required_fuel field to continue execution

cemonem:
Resumables might be supplied with fuel that is too little to continue execution. This addition suggests a simple change to the Resumable API to inform the user with the minimum required fuel amount to continue execution, and also supplies a method to inspect/modify the remaining fuel within the resumable.

### Checks

- Using Nix
  - [x] Ran `nix fmt`
- Using Rust tooling
  - [x] Ran `cargo fmt`
  - [x] Ran `cargo test`
  - [x] Ran `cargo check`
  - [x] Ran `cargo build`
  - [x] Ran `cargo doc`

### DIFF
--- a/src/execution/error.rs
+++ b/src/execution/error.rs
@@ -14,7 +14,6 @@ pub enum RuntimeError {
     InvalidImportType,
     UnknownImport,
     MoreThanOneMemory,
-    OutOfFuel,
 }
 
 impl Display for RuntimeError {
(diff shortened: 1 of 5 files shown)


florianhartung (changes requested):

src/execution/error.rs
```diff
@@ -14,7 +14,7 @@ pub enum RuntimeError {
     InvalidImportType,
     UnknownImport,
     MoreThanOneMemory,
-    OutOfFuel,
+    OutOfFuel { required_fuel: u32 },
```
wucke13: We should not make this a public error, the user will never see it:

either there is no fuel enabled, or an out-of-fuel event will trigger a resumable to be returned to the user.

wucke13 (approved):

PR #270, opened 2025-10-17
move resumable methods into store, create resumables without invoking them

cemonem:
### Pull Request Overview

ResumableRef objects require the owning RuntimeInstance to be operated on. Moving them back to Store provides cleaner interface along with create_resumable method for cases when we want to create one without invoking it.

### Checks

- Using Nix
  - [x] Ran `nix fmt`
- Using Rust tooling
  - [x] Ran `cargo fmt`
  - [x] Ran `cargo test`
  - [x] Ran `cargo check`
  - [x] Ran `cargo build`
  - [x] Ran `cargo doc`

### DIFF
--- a/src/execution/store.rs
+++ b/src/execution/store.rs
@@ -1,3 +1,5 @@
+use core::mem;
+
 use crate::core::indices::TypeIdx;
 use crate::core::reader::span::Span;
 use crate::core::reader::types::data::{DataModeActive, DataSegment};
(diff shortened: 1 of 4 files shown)


src/execution/store.rs
```diff
@@ -557,65 +562,170 @@ impl<'b, T> Store<'b, T> {
...
+            ResumableRef::Fresh(FreshResumableRef {
+                func_addr,
+                params,
+                maybe_fuel,
+            }) => {
+                // Err can never happen here
+                let func_inst = self
+                    .functions
+                    .get(func_addr)
+                    .ok_or(RuntimeError::FunctionNotFound)?;
```
florianhartung:
```suggestion
                let func_inst = self
                    .functions
                    .get(func_addr)
                    .expect("func addrs to always be valid if the correct store is used")
```

Let's panic for now. The planned `Stored<T>` object will fix these issues.

florianhartung (changes requested):

florianhartung (approved):

Issue #157, opened 2025-04-09
Newtypes for addr types


florianhartung:
The Wasm spec uses addr types for most indices: https://webassembly.github.io/spec/core/exec/runtime.html#addresses
In the future we definitely want to use Rust's newtypes for such indices and addresses for safety and to avoid confusion.

PR #281, opened 2025-10-27
Address Types

florianhartung:
~~_This PR's branch is based on top of [Embedder API #285](https://github.com/DLR-FT/wasm-interpreter/pull/285)'s branch. Therefore, the first **4 commits** are not part of this PR._~~

A rework of the address types and address spaces in the `Store`.

First, this PR includes various fixes for tests, which previously relied on internals of the `Store`. For example, some tests would perform an access like `store.globals[0]` to read the global with the global address 0. With these fixes done, the commits adding the actual `*Addr` newtypes become much smaller.

The newtypes are added in 7 small commits, where each commit adds a new `*Addr` type and an address space in the `Store`.
Instances in the store are now stored as `AddrVec<FuncAddr, FuncInst>`, where `AddrVec` is a newtype around a `Vec` with a slightly different interface. 

### Checks

- Using Nix
  - [x] Ran `nix fmt`
- Using Rust tooling
  - [ ] Ran `cargo fmt`
  - [ ] Ran `cargo test`
  - [ ] Ran `cargo check`
  - [ ] Ran `cargo build`
  - [ ] Ran `cargo doc`

### Github Issue

Closes https://github.com/DLR-FT/wasm-interpreter/issues/157


cemonem:
Lgtm, but the only minutiae I could point out is

- the naming `AddrVec` (these hold instances, not `addrs`, and the `ModuleInsts` *do* hold vectors of addrs)
- the possibility of not specifying the `Addr` type in `AddrVec/InstVec` with an `Inst` trait:
```
trait Inst {
    type Addr: Addr;
}

struct InstVec<I: Inst> {
    inner: Vec<I>
}

impl<I: Inst> InstVec<I> {
    pub fn get(&self, addr: I::Addr) -> &I {
        self.inner.get(addr.into_inner()).expect("addrs to always be valid")
    }
}
```
- we could just hold the `module_addr` of the last defined module in `testsuite::run.rs` instead of all `modules` but this part requires heavy refactor in the future anyway, so its ok by me.

cemonem (approved):

PR #288, opened 2025-11-10
feat: config methods for fuel for each instruction

cemonem:
This commit adds `Config::*_flat_cost()` methods in which the user can supply the fuel to be deducted when a particular instruction is fetched. In order to make sure simple implementations that match the instructions into constants are folded into a constant themselves in the interpreter loop the method is called in each match arm in the interpreter fetch loop instead of once at the beginning.

### Checks

- Using Nix
  - [x] Ran `nix fmt`
- Using Rust tooling
  - [x] Ran `cargo fmt`
  - [x] Ran `cargo test`
  - [x] Ran `cargo check`
  - [x] Ran `cargo build`
  - [x] Ran `cargo doc`

### DIFF
--- a/src/execution/config.rs
+++ b/src/execution/config.rs
@@ -49,6 +49,47 @@ pub trait Config {
     fn get_fd_extension_flat_cost(_instr: u32) -> u32 {
         1
     }
+
+    /// Amount of fuel to be deducted per element of a single byte instruction `instr` that executes in asymptotically
+    /// linear time with respect to one of the values it pops from the stack.
+    ///
+    /// In Wasm 2.0 specification, this applies to the following instructions:
+    /// - `MEMORY.GROW` of type `[n: i32] -> [i32]`
+    ///
+    /// The cost of the instruction is calculated as `cost := get_flat_cost(instr) + n*get_cost_per_element(instr)`.
+    /// where `n` is the stack value marked in the instruction type signature above. Other instructions and bytes that
+    /// do not correspond to any instruction are ignored.
+    // It must always be checked that the calls to this method fold into a constant if it is just a match statement that
+    // yields constants.
+    #[inline(always)]
+    fn get_cost_per_element(_instr: u8) -> u32 {
+        0
+    }
+
+    /// Amount of fuel to be deducted per element of a  multi-byte instruction that starts with the byte 0xFC,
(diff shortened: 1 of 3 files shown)


src/execution/config.rs
```diff
@@ -18,6 +18,40 @@ pub trait Config {
...
+
+    /// Amount of fuel to be deducted when a single byte `instr` is hit. The cost corresponding to `UNREACHABLE`
+    /// instruction and other bytes that do not correspond to any Wasm instruction are ignored.
+    // It must always be checked that the calls to this method fold into a constant if its just a match statement that
+    // yields constants.
+    #[allow(unused_variables)]
+    #[inline(always)]
+    fn get_flat_cost(instr: u8) -> u32 {
+        1
+    }
```
florianhartung:
```suggestion
    #[inline(always)]
    fn get_flat_cost(_instr: u8) -> u32 {
        1
    }
```

src/execution/interpreter_loop.rs
```diff
@@ -99,9 +104,11 @@ pub(super) fn run<T: Config>(
 
         match first_instr_byte {
             NOP => {
+                decrement_fuel!(T::get_flat_cost(NOP), 1);
                 trace!("Instruction: NOP");
             }
             END => {
+                decrement_fuel!(T::get_flat_cost(END), 1);
```
florianhartung: Should a function's `END` really consume fuel? Technically it is not an instruction..

src/execution/interpreter_loop.rs
```diff
@@ -2654,6 +2862,7 @@ pub(super) fn run<T: Config>(
                 use crate::core::reader::types::opcode::fd_extensions::*;
                 match second_instr {
                     V128_LOAD => {
+                        decrement_fuel!(T::get_fd_extension_flat_cost(V128_LOAD), 5);
```
florianhartung: I don't think rewind=5 is correct because multi-byte instructions are encoded as a u8 + variable-length u32.

src/execution/interpreter_loop.rs
cemonem: Yes, i should remove this for now. This instruction is also skipped depending on whether a branch is taken or not.

Issue #293, opened 2025-11-14
return the remaining amount of fuel even if execution succeeds


cemonem:
Currently the remaining amount of fuel is only accessible if execution terminates due to insufficient fuel. However the remaining amount of fuel is valuable information even if the execution terminates.

florianhartung:
But how can we make it so that this also works for the execution of the start function?

cemonem:
> I think we can just make `RunState` include this information:
> 
> pub enum RunState {
>     Finished {
>         values: Vec<Value>,
>         maybe_remaining_fuel: Option<u32>,
>     },
>     Resumable {
>         resumable_ref: ResumableRef,
>         required_fuel: NonZeroU32,
>     },
> }

Yeah I was thinking of something like this. For the start function, I think we can return a `FueledInstantiation{module_addr: ModuleAddr, remaining_fuel: Option<u32>}` instead of `ModuleAddr` in module_instantiate and call it a day.

PR #294, opened 2025-11-14
feat: return remaining fuel in finished resumable

cemonem:
The user might want to query how much fuel remained, even if the resumable terminated completely.

Related to #293 

### Checks

- Using Nix
  - [x] Ran `nix fmt`
- Using Rust tooling
  - [x] Ran `cargo fmt`
  - [x] Ran `cargo test`
  - [x] Ran `cargo check`
  - [x] Ran `cargo build`
  - [x] Ran `cargo doc`

### DIFF
--- a/src/execution/linker.rs
+++ b/src/execution/linker.rs
@@ -4,7 +4,9 @@ use alloc::{
     vec::Vec,
 };
 
-use crate::{addrs::ModuleAddr, ExternVal, RuntimeError, Store, ValidationInfo};
+use crate::{
+    addrs::ModuleAddr, store::InstantiationOutcome, ExternVal, RuntimeError, Store, ValidationInfo,
+};
 
 use super::config::Config;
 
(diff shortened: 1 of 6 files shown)


florianhartung (approved):

src/execution/resumable.rs
```diff
@@ -72,17 +72,32 @@ impl Drop for InvokedResumableRef {
...
     Resumable {
         resumable_ref: ResumableRef,
         required_fuel: NonZeroU32,
     },
 }
 
+/// Represents a successful, possibly fueled instantiation of a module. `module_addr` contains the store address of the
+/// module that has successfully instantiated. `maybe_remaining_fuel` contains the `Some(remaining_fuel)` if
+/// instantiation was fuel-metered and `None` otherwise.
+pub struct FueledInstantiation {
```
florianhartung: How about `InstantiationResult` instead?
Or `InstantiationOutcome` so it cannot be confused with `Result` types?

PR #340, opened 2026-04-17
feat: validation instruction hook

cemonem:
### Pull Request Overview
#323 
Simple hook that gets called for every instruction when during validation of code section. 

### TODO or Help Wanted

- I am not sure this is enough or maybe too much, the hook doesn't know about which function it was called in, for example. On the other hand, i'm not sure just being able to scan instructions necessitates a hook, someone could write a crude pass to eliminate binaries with unwanted instructions without even using our codebase. On the other hand, if they want to take advantage of the context of validation, i don't think `(bytecode, pc)` tuple is enough.
- <del>Namespace clashes when i move ValidationConfig into its own `validation/config.rs`</del> (I renamed the module as `validation_config` but the path name is `validation/config.rs`).

### Checks

- Using Nix
  - [x] Ran `nix fmt`
- Using Rust tooling
  - [x] Ran `cargo fmt`
  - [ ] Ran `cargo test`
  - [ ] Ran `cargo check`
  - [x] Ran `cargo build`
  - [ ] Ran `cargo doc`

### DIFF
--- a/crates/benchmark/benches/general_purpose.rs
+++ b/crates/benchmark/benches/general_purpose.rs
@@ -53,7 +53,7 @@ macro_rules! bench_wasm {
             let wasm_bytes = $wasm_bytes;
 
             // Our interpreter
-            let our_validation_info = decode_and_validate(&wasm_bytes).unwrap();
+            let our_validation_info = decode_and_validate(&wasm_bytes, ()).unwrap();
             struct UserData;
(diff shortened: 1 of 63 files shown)


florianhartung (approved):

florianhartung:
Closes https://github.com/DLR-FT/wasm-interpreter/issues/323

PR #341, opened 2026-04-22
Fix Bug Where the PC is Reset to an Unknown Location When Out Of Fuel in Fill/Init Instructions

florianhartung:
This also adds a single test case for the `memory.fill` instruction specifically.

### Checks

- Using Nix
  - [x] Ran `nix fmt`
- Using Rust tooling
  - [ ] Ran `cargo fmt`
  - [ ] Ran `cargo test`
  - [ ] Ran `cargo check`
  - [ ] Ran `cargo build`
  - [ ] Ran `cargo doc`


florianhartung:
It might make sense to add more tests, which test for fuel-related control flow paths. This bug has not been found for quite a while, because we don't have a lot of fuel tests.

Or we could just execute the entire spec testsuite instruction-by-instruction..

cemonem (approved):

PR #366, opened 2026-07-09
Fix FixedCapacityVec

skade:
`FixedCapacityVec::pop_into_slice` has the potential for a memory violation due to an improper bounds check. Instead of checking against the number of contained elements, it checks against the length of the container.

This case is caught by running an offending example in debug mode, e.g. in test mode.

The test is included as a single patch.

Another patch consistently makes sure that capacity checks are using the `capacity()` accessor and `elements.len()`.

# Open Questions

Few.

# Benchmark Results

Benchmarks should now complete.

# References

Now all legal.

# Checks

I added a test.

- Using Nix
  - [ ] Ran `nix fmt`
- Using Rust tooling
  - [x] Ran `cargo fmt`
  - [x] Ran `cargo test`
  - [ ] Ran `cargo check`
  - [x] Ran `cargo build`

### DIFF
--- a/src/core/fixed_capacity_vec.rs
+++ b/src/core/fixed_capacity_vec.rs
@@ -340,3 +340,18 @@ unsafe fn slice_assume_init<T>(slice: &[MaybeUninit<T>]) -> &[T] {
     // reference and thus guaranteed to be valid for reads.
     unsafe { &*(slice as *const _ as *const [T]) }
 }
+
+#[cfg(test)]
+mod test {
+    use super::*;
+
+    #[test]
+    fn pop_into_slice_rejects_more_than_len() {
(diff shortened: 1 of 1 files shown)


florianhartung:
First of all, thanks for your contribution.

The patches look good to me.

However CI is still complaining because the commits are not signed off. Could you please sign off all three commits as described [here](https://github.com/DLR-FT/wasm-interpreter/pull/366/checks?check_run_id=86218988731) and run `cargo fmt`.

Also we prefer the conventional commit message format to be used, however this is not a blocker.

florianhartung (approved):

wucke13:
@skade Thanks a lot!

PR #376, opened 2026-07-23
Reduce Amount of Allocations

florianhartung:
# Open Questions / TODOs

# Benchmark Results

Coremark Results:

On main I get 311 and 312. On this branch I get 344 and 348.

I expect that the optimized br_table execution is responsible for this. Maybe it's in a hot loop somewhere. The coremark benchmark definitly contains ~10 br.table instructions

# References

# Checks

- Using Nix
  - [x] Ran `nix fmt`
- Using Rust tooling
  - [ ] Ran `cargo fmt`
  - [ ] Ran `cargo test`
  - [ ] Ran `cargo check`
  - [ ] Ran `cargo build`


src/execution/instructions/const_interpreter_loop.rs
```diff
@@ -92,15 +92,26 @@ pub(crate) unsafe fn run_const_span<T: Config>(
...
     module: ModuleAddr,
     store: &Store<T>,
+    maybe_reusable_stack: &mut Option<Stack>,
 ) -> Result<Option<Value>, RuntimeError> {
     let mut wasm = WasmDecoder::new(wasm);
 
     wasm.move_start_to(*span).unwrap_validated();
 
-    let mut stack = Stack::new::<T>(Vec::new(), &FuncType::new_empty(), &[])?;
+    // If there is a stack to use, clear and reinitialize it. Otherwise create a new stack.
```
cemonem, line 94: in a future pr, we can remove this stack all together and have a flag if we're supposed to have more than 1 item in the stack (that we pretend to have) since all constexpr instructions are already values by themselves

florianhartung: AFAIK we discussed this topic already. For now, I'd like to keep the stack because it allows us to easily implement the [Extended Const Expressions](https://github.com/WebAssembly/extended-const) proposal (which is part of Wasm 3)

florianhartung: However, the value stack may be untyped for constant expressions. https://github.com/DLR-FT/wasm-interpreter/issues/394 combined with https://github.com/DLR-FT/wasm-interpreter/issues/389 may therefore result in quite performant code

PR #395, opened 2026-07-29
Remove Logging

florianhartung:
# Open Questions / TODOs

What do we do about REQ-15 "Allow log dependency". Do we just remove it? Mark it as outdated?

# Benchmark Results

# References

# Checks

- Using Nix
  - [x] Ran `nix fmt`
- Using Rust tooling
  - [ ] Ran `cargo fmt`
  - [ ] Ran `cargo test`
  - [ ] Ran `cargo check`
  - [ ] Ran `cargo build`


src/execution/runtime_structure/addresses.rs
wucke13, line 1: We should keep the display implementations for these things, as they are externally handled.

src/core/structure/instructions.rs
wucke13, line 1: I do think these are useful, but we should/can move them into an auxillary crate

wucke13: Covered by #409

PR #403, opened 2026-08-03
Make MemInst::grow Return The Previous Length

florianhartung:
Because growing with shared linear memories is defined a little different in the spec, we can no longer simply read the previous length in the mem.grow instruction handler. Instead the memory instance should return in. 

# Open Questions / TODOs

# Benchmark Results

# References

# Checks

- Using Nix
  - [x] Ran `nix fmt`
- Using Rust tooling
  - [ ] Ran `cargo fmt`
  - [ ] Ran `cargo test`
  - [ ] Ran `cargo check`
  - [ ] Ran `cargo build`


src/execution/runtime_structure/memory_instances/mod.rs
```diff
@@ -25,34 +27,35 @@ impl MemInst {
             return Err(RuntimeError::MemoryGrowOverflowed);
         };
 
+        // We limit the number of pages to 2^16 because otherwise addresses for the next page's
+        // contents would exceed 2^32 - 1.
+        if new_len_pages > 2u32.pow(16) {
```
wucke13: I think this should be `1<<16 - 1` or use `>=` as comparator

PR #404, opened 2026-08-03
Limit Memory Size To u32 Bytes & Use u32 Indices

florianhartung:
# Open Questions / TODOs

# Benchmark Results

Two Coremark Runs:
- Before: 276.41, 276.13
- After: 283.35, 290.55

Interpretation: u32 is only half as big as a usize as a function argument. Maybe less data to move between functions sped up some memory instructions.

# References

First commit is from https://github.com/DLR-FT/wasm-interpreter/pull/403.

# Checks

- Using Nix
  - [x] Ran `nix fmt`
- Using Rust tooling
  - [ ] Ran `cargo fmt`
  - [ ] Ran `cargo test`
  - [ ] Ran `cargo check`
  - [ ] Ran `cargo build`


wucke13:
From a data oriented programming perspective, this is the right move. However, this brings us one step away from having 64 bit linear memory support. I would argue that, having 32 bit host 64 bit guest is nonsense; and thus if we use `usize`, we have already native support for 64 bit linear memories on all relevant targets, which with this we do loose. Then again, I do think that 32 bit is more than fine for most applications. What are your thoughts @florianhartung ?

florianhartung:
@wucke13 I agree. I've changed the implementation back to using using usizes and kept the other improvements made along the way. Also I added user-customizable limits for memory and table sizes.

